PR #44, opened 2018-03-09
fix#42: Add PingSettings

patrickelectric:


patrickelectric:
This will be done in cpp with a singleton.

PR #81, opened 2018-04-11
Fix#78: Enable manual serial connection 

patrickelectric:


qml/MainMenu.qml
```diff
@@ -31,6 +42,7 @@ Item {
                             ? "5:-:-"
                             : (conntype.currentIndex + 2).toString() + ":" + first + ":" + second
 
+        ping.stopDetector()
```
jaxxzer: move this to ping object connectlink logic

PR #83, opened 2018-04-16
Fix#78: Open again, github problems..

patrickelectric:


patrickelectric:
@jaxxzer modifications done.

PR #114, opened 2018-05-22
Use ping1DNamespace

patrickelectric:
Depends of bluerobotics/ping-protocol-cpp#33


patrickelectric:
This is having some problems to build, work around this is necessary...

jaxxzer:
Another reason for doing, https://github.com/bluerobotics/ping-protocol-cpp/issues/19
it will reveal the functional changes in a patch like this.

PR #115, opened 2018-05-23
Fix problems with focus and text fields

jaxxzer:


jaxxzer:
Fix #116

patrickelectric:
Looks good, just need to correct the commit prefix, `PingTextField:`

PR #117, opened 2018-05-23
.pro: force gcc 8

jaxxzer:


patrickelectric:
Why is this necessary ?

PR #118, opened 2018-05-26
more ping interface implementations

### LINKED ISSUE #92: Implement entire ping software interface in UI.
Add range settings and auto/manual mode

jaxxzer:
Closes #92

### DIFF
--- a/qml/MainMenu.qml
+++ b/qml/MainMenu.qml
@@ -110,6 +110,50 @@ Item {
                         ping.gain_index = currentIndex
                     }
                 }
+                Text {
+                    text: "Start/Stop (mm):"
+                    color: Style.textColor
+                }
+
+                PingTextField {
+                    id: startLength
+                    text: ""
+                    Layout.columnSpan: 2
+                    Layout.fillWidth: true
+                    onEditingFinished: {
+                        if (parseInt(text)) {
+                            ping.start_mm = parseInt(text)
+                        }
+                    }
+                    Connections {
+                        target: ping
+                        onLengthMmUpdate: {
+                            if (!startLength.focus) {
+                                startLength.text = ping.start_mm
+                            }
+                        }
+                    }
+                }
+
+                PingTextField {
+                    id: totalLength
+                    text: ""
+                    Layout.columnSpan: 2
+                    Layout.fillWidth: true
+                    onEditingFinished: {
+                        if (parseInt(text)) {
+                            ping.length_mm = parseInt(text)
+                        }
+                    }
+                    Connections {
+                        target: ping
+                        onLengthMmUpdate: {
+                            if (!totalLength.focus) {
+                                totalLength.text = ping.length_mm
+                            }
+                        }
+                    }
+                }
             }
         }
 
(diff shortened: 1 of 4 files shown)


qml/MainMenu.qml
```diff
@@ -90,16 +90,15 @@ Item {
                 CheckBox {
                     id: autoGainChB
                     text: "Auto Gain"
-                    checked: ping.modeAuto
+                    checked: ping.mode_auto
                     onCheckedChanged: {
-                        //setEchosounderAuto(checked)
-                        ping.request(1102)
+                        ping.mode_auto = checked
```
patrickelectric: You need to request the mode, otherwise will not be updated in the interface if something goes wrong.

qml/MainMenu.qml
```diff
@@ -108,7 +107,51 @@ Item {
...
+                }
+
+                PingTextField {
+                    id: totalLength
+                    text: ""
+                    Layout.columnSpan: 2
+                    Layout.fillWidth: true
+                    onEditingFinished: {
+                        if (parseInt(text)) {
+                            ping.length_mm = parseInt(text)
```
patrickelectric: I edited the field and ping stop to response.

qml/PingStatus.qml
```diff
@@ -16,7 +16,7 @@ Item {
         color: "black"
         opacity: 0.75
         height: innerCol.height
-        width: innerCol.width*1.2
+        width: parent.width * 0.2
```
patrickelectric: This is huge. 
screenshot from 2018-05-28 10-34-27

src/sensor/ping.cpp
```diff
@@ -175,6 +175,30 @@ void Ping::handleMessage(PingMessage msg)
     }
     break;
 
+    case Ping1DNamespace::Range: {
+        ping_msg_ping1D_range m(msg);
+        _start_mm = m.start_mm();
+        _length_mm = m.length_mm();
+        qDebug() << "got range" << _start_mm << _length_mm;
```
patrickelectric: qCDebug

qml/PingStatus.qml
jaxxzer: So? it keeps the numbers from all jumping around when you get 2 billion for length or something when there is bugs.

src/sensor/ping.cpp
jaxxzer: ok

qml/PingStatus.qml
patrickelectric: > all jumping around when you get 2 billion for length

If the number can go this big, use toExponential(5) or something like it to short the value.

jaxxzer: if the numbers are changing at 30Hz, decimal will be able to see there is a problem with human eyes.
toExponential maybe not

jaxxzer: fixed it all lmk if you like it

src/sensor/ping.cpp
jaxxzer: took it out, did not mean to commit

jaxxzer:
Closes #92

Issue #22, opened 2018-03-06
Relative sizing of UI elements


jaxxzer:
*Nice to have*
Adjust UI sizing/placement based on window size/ screen size.

PR #126, opened 2018-06-07
Use InfoPage in fill mode

patrickelectric:
This is also a functionality test, showing how dynamic and fluid layouts can be done better with "fill" item types.
There is also a video in slack.


patrickelectric:
This is an alternative solution for #22

patrickelectric:
Fixed.

PR #150, opened 2018-06-19
Waterfall: Use NoFormatConversion with fromImage

patrickelectric:
From: http://lists.qt-project.org/pipermail/interest/2015-October/019264.html
> Converting that to a QPixmap will entail a QImage::convertToFormat() to map the image into a system specific pixel format, usually QImage::RGB32 or
> QImage::ARGB32_Premultiplied, but any given QPA backend may override that with an other format if that is more compatible with its raster backingstore.

This approach reduce CPU usage in 27%


patrickelectric:
The right window shows the result of this PR.
screenshot from 2018-06-19 16-32-17

PR #202, opened 2018-06-28
Depth axis

### LINKED ISSUE #40: depth axis scale/reading

jaxxzer:
fix #40

### DIFF
--- a/qml/DepthAxis.qml
+++ b/qml/DepthAxis.qml
@@ -1,0 +1,147 @@
+import QtQuick 2.7
+import QtQuick.Controls 2.2
+import QtQuick.Layouts 1.3
+
+Item {
+    id: root
+
+    // these are actually all in meters right now
+    property var start_mm: 0 // The depth at the top edge of the item
+    property var end_mm: 0 // The depth at the bottom edge of the item
+    property var length_mm: end_mm - start_mm
+    property var depth_mm: 0 // The range to target (bottom) currently reported by the ping
+
+    property var validIncrements: [0.1, 0.25, 0.5, 1, 2, 5, 10]
+
+    property int numTicks: 10
+    property int maxTicks: 10
+    property var increment: 1 // increment to use between tick marks
+
+    property var color: "white"
+
+    function getIncrement() {
+        var inc;
+        for (var i = 0; i < validIncrements.length; i++) {
+            inc = validIncrements[i];
+            if (length_mm / inc <= maxTicks) {
+                break;
+            }
+        }
+        return inc;
+    }
+
+    // Update increment size and number of ticks to display
+    function recalc() {
+        increment = getIncrement()
+        numTicks = length_mm / increment + 1
+    }
+
+    onStart_mmChanged: {
+        recalc()
+    }
+
+    onEnd_mmChanged: {
+        recalc()
+    }
+
+    // The pixel distance between ticks at regular intervals
+    function nominalHeight() {
+        return increment * parent.height / length_mm
+    }
+
+    // The remaining height between the penultimate and last tick mark at the bottom of the item
+    function remainingHeight() {
+        return parent.height - ((numTicks-1) * nominalHeight())
+    }
+
+    Component {
+        id: tickMark
+        RowLayout {
+            anchors.right: parent.right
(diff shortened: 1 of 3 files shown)


qml/DepthAxis.qml
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    Image {
+        anchors.right: parent.right
+        source: "/icons/depth_indicator.svg"
+        height: 20
+        width: 20
+        y: (parent.height / length_mm) * depth_mm - height / 2
+    }
```
patrickelectric: For simple things like triangle, or square, you can use canvas:
```cpp
    Canvas {
        id: depthPointer
        y: (parent.height / length_mm) * depth_mm - height / 2
        z: root.z + 1
        width: 20
        height: width
        onWidthChanged: requestPaint()
        onPaint: {
            print(width, height)
            var ctx = getContext("2d")
            ctx.fillStyle = "red"
            ctx.beginPath()
            ctx.moveTo(width, 0)
            ctx.lineTo(0, height/2)
            ctx.lineTo(width, height)
            ctx.fill()
        }
    }
```

qml/DepthAxis.qml
```diff
@@ -0,0 +1,128 @@
+import QtQuick 2.7
+import QtQuick.Controls 2.2
+import QtQuick.Layouts 1.3
+
+import Logger 1.0
```
patrickelectric: Why logger is necessary here ?

qml/DepthAxis.qml
```diff
@@ -0,0 +1,128 @@
...
+            height: lastIndex ? remainingHeight() : nominalHeight()
+            Layout.fillHeight: true
+            Layout.alignment: Qt.AlignRight | Qt.AlignBottom
+
+            property var lastIndex: (index + 1) == numTicks
+
+            Text {
+                id: label
+                //anchors.fill:parent // this text object is not anchored
+                //Layout.fillHeight: true
```
patrickelectric: This code is for debug or can be deleted ?

jaxxzer: It is a reminder for anyone who adjusts it in the future. It is a little bit nonconventional how I got it to look right I think.

jaxxzer: I can try to rework it

qml/DepthAxis.qml
jaxxzer: copy paste

qml/DepthAxis.qml
```diff
@@ -0,0 +1,128 @@
...
+    property var length_mm: end_mm - start_mm
+    property var depth_mm: 0
+
+    property var validIncrements: [0.1, 0.25, 0.5, 1, 2, 5, 10]
+
+    property int numTicks: 10
+    property int maxTicks: 10
+    property var increment: 1
+
+    property var color: "white"
```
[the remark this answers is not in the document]
jaxxzer: where is Style.color determined?

qml/DepthAxis.qml
jaxxzer: That would require thought and effort

patrickelectric: This code already do the red triangle. You just need to copy and paste :)

qml/DepthAxis.qml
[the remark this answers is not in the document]
jaxxzer: The parent can set the color property of this item to Style.color if necessary. Style.color is giving me black which is not very good to see for most themes.

qml/DepthAxis.qml
jaxxzer: thanks for doing the hard work for me

qml/DepthAxis.qml
jaxxzer: There are some bugs with what I expect to be the reasonable thing to do (bind text anchors to parent, use horizontal and verticalAlignment). It hangs the program when trying to use these. I left the comments just to explain the situation.

jaxxzer:
It's fixed. I think we should try some real life tests with this version and change it subsequently if we decide it's not right.

jaxxzer:
@patrickelectric delete branches after merge please.

PR #206, opened 2018-07-03
Settings manager: Create a class to manage all shared settings

### LINKED ISSUE #42: Create a singleton cpp class to handle settings

### LINKED ISSUE #205: Disable values inputs check in debug mode.

patrickelectric:
Fix #205 
Will help #199 
Fix #42

### DIFF
--- a/src/logger.cpp
+++ b/src/logger.cpp
@@ -1,5 +1,6 @@
 #include "filemanager.h"
 #include "logger.h"
+#include "settingsmanager.h"
 
 #include <iostream>
 #include <QDebug>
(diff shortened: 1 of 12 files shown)


src/logger.cpp
```diff
@@ -42,7 +34,7 @@ void Logger::installHandler()
     originalHandler = qInstallMessageHandler(handleMessage); // This function returns the previous message handler
 
     if (qEnvironmentVariableIsEmpty("QT_MESSAGE_PATTERN")) {
-        qSetMessagePattern(QStringLiteral("%{time [hh:mm:ss.zzz]} %{message}"));
```
jaxxzer: Was adding this hunk intentional? What does this do?

src/logger.h
```diff
@@ -115,7 +114,6 @@ class Logger : public QObject
     QFile _file;
     QTextStream _fileStream;
     QStringList _registeredCategories;
-    QSettings _settings;
```
jaxxzer: Can we assign SettingsManager::self()->settings() to a member _settings, and use that everywhere for better readability?

src/logger.cpp
patrickelectric: Ops, my bad, I was using a personal pattern while debugging.

src/logger.h
patrickelectric: Yep, Will do that

PR #208, opened 2018-07-04
Add board temperature and voltage

jaxxzer:


jaxxzer:
The pcb temperature comes back.. but also a nack. I meant to make a note of this sorry.

PR #218, opened 2018-07-19
LogListModel corrections

patrickelectric:
- Move const variables that belong to model struct to private variables.
  - They are defined only once
  - They define how the entire model will work, makes more sense to move it to header


jaxxzer:
Good catch. Is there a performance increase after this patch (w high rate logging)?

patrickelectric:
Yes, the number of memory allocated and unallocated decrease with this patch, besides the processing time.

PR #234, opened 2018-07-30
[WIP] Add CSV parser

### LINKED ISSUE #131: Convert .bin files to .csv
Add a tool to export log files to csv

patrickelectric:
Fix #131 
Needs [#61](https://github.com/bluerobotics/ping-protocol-cpp/pull/61)


patrickelectric:
Needs @brianhBR approval.

src/sensor/ping.cpp
```diff
@@ -444,3 +447,92 @@ void Ping::writeMessage(const PingMessage &msg)
...
+        // This will avoid redundant emition
+        if(lastPercentage != percentage) {
+            lastPercentage = percentage;
+            emit csvProgress(percentage);
+        }
+
+        //Check for valid checksum and id
+        ping_msg_ping1D_empty ping_message{(uint8_t*)(packages[i].data()), (uint16_t)packages[i].length()};
+        if(!ping_message.verifyChecksum()) {
+            continue;
```
jaxxzer: we should warn here.

jaxxzer:
I'm wondering what's the most useful thing here. For example, if we receive distance info from two different messages, the data lands in two different files. Perhaps we should stick to only pertinent information (confidence, range, profile) and interleave this data from all messages into one output file. Meanwhile, we can forgo exporting things like firmware version, voltage, scan range etc. This consideration should not hold up this PR, however.

patrickelectric:
> I'm wondering what's the most useful thing here. For example, if we receive distance info from two different messages, the data lands in two different files. Perhaps we should stick to only pertinent information (confidence, range, profile) and interleave this data from all messages into one output file. Meanwhile, we can forgo exporting things like firmware version, voltage, scan range etc. This consideration should not hold up this PR, however.

I would like to add each variable in a different tab in excel, and in the end only one file with a bunch of tabs. But this will need second party libraries or new classes to handle everything. This is a simple and good first version.

> Why are all the function mappings returning floats?
Is it because the functions in the map must all have the same type?

Exactly:
https://github.com/bluerobotics/ping-protocol-cpp/pull/61/files#diff-ebcbbb8c54e652539da8ea1febd4bf6eR34

jaxxzer:
Add timestamps or open an issue for it after merge.

patrickelectric:
Closing due to inactivity.

PR #278, opened 2018-09-04
Improve travis script

patrickelectric:


.travis.yml
```diff
@@ -124,70 +124,35 @@ script:
...
 - export DEPLOY_FILE="";
 - if [ -z $TEST ]; then
+    ls /tmp/pingviewer*;
     wget -c https://github.com/probonopd/uploadtool/raw/master/upload.sh;
   fi
 - if [ ! -z $TRAVIS_BRANCH ] && [ "$TRAVIS_BRANCH" != "master" ] ; then
     export UPLOADTOOL_SUFFIX=$TRAVIS_BRANCH;
   fi
 - if [[ "${TRAVIS_OS_NAME}" = "linux" && -z $TEST ]]; then
     bash ./upload.sh ./pingviewer*.AppImage;
```
jaxxzer: `./pingviewer*.AppImage` -> `/tmp/pingviewer*.AppImage`

better yet, move DEPLOY_FILE above this and use that. Same as below?

.travis.yml
```diff
@@ -124,70 +124,35 @@ script:
...
-    mv pingviewer.dmg pingviewer-${QT_CONFIG}.dmg;
+    ./tools/compile.sh || travis_terminate 1;
   fi
 
 after_success:
 # Upload files in GitHub
 - export DEPLOY_FILE="";
 - if [ -z $TEST ]; then
+    ls /tmp/pingviewer*;
     wget -c https://github.com/probonopd/uploadtool/raw/master/upload.sh;
```
jaxxzer: Should we fork this? What if he changes it and breaks our build?

patrickelectric: For small projects such this one I would say yes. But probonopd and theassassin are pretty popular in deployment tools, I trusty them. Besides that, linuxdeployqt and uploadtool both are tested and deployed by it's own release, so if they do not work, they will not be available to download :)

.travis.yml
patrickelectric: Done.

PR #288, opened 2018-09-10
Add first tests of the interface

patrickelectric:
Needs https://github.com/bluerobotics/ping-viewer/pull/287

### DIFF
--- a/src/test.cpp
+++ b/src/test.cpp
@@ -1,0 +1,76 @@
+#define private public
+#define protected public
+
+#include <QApplication>
+#include <QQmlApplicationEngine>
+#include <QQmlContext>
+#include <QQmlEngine>
+#include <QQuickStyle>
+#include <QDebug>
+#include <QRegularExpression>
+
+#include "abstractlink.h"
+#include "filemanager.h"
+#include "linkconfiguration.h"
+#include "logger.h"
+#include "ping.h"
+#include "settingsmanager.h"
+#include "util.h"
+#include "waterfall.h"
+
+#include "test.h"
+
+// Register message enums to qml
+#include "pingmessage/pingmessage.h"
+
+void Test::initTestCase()
+{
+    FileManager::self();
+}
+
+void Test::fileManager()
+{
+    auto fileManager = FileManager::self();
+    // Check if folders exist and if it's ok
+    for(const auto& folder : fileManager->folderMap) {
+        QVERIFY2(folder->dir.exists(),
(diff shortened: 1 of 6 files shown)


src/test.cpp
```diff
@@ -132,4 +132,43 @@ void Test::settingsManager()
              qPrintable(QString("Distance scalar in meters is wrong: %1").arg(scalar)));
 }
 
+void Test::waterfallGradient()
+{
+    QVector<QColor> colorList = {Qt::black, Qt::white};
+    // Check name and validation
+    auto testName = QString("Test");
```
jaxxzer: `"Waterfall gradient test"`?

src/test.cpp
```diff
@@ -132,4 +132,43 @@ void Test::settingsManager()
...
+    auto gradient = WaterfallGradient(testName, colorList);
+    QVERIFY2(gradient.isOk(), qPrintable("Basic gradient is not ok!"));
+    QVERIFY2(gradient.name().contains(testName),
+             qPrintable(QString("Name does not match: %1 != %2.").arg(gradient.name(), testName)));
```
jaxxzer: i see...

jaxxzer:
Looks ok, thanks for the work in this area!

My only concern at the moment is putting tests for all our different objects into a single test class/file.

It seems to me that we should make a different cpp for each class that we are testing. We could have a base virtual test class, and subclass it for filemanager, settingsmanager etc.

What do you think? Is there some reason that you are putting everything in one cpp?

src/test.cpp
patrickelectric: It's only a string to create a valid waterfall, the only thing that is necessary here is not to be empty.

patrickelectric:
You can use they in a single file or split, right now for just small tests a file for each one appears unnecessary to me, each test is +-30 lines and we only have four. In the future with more advanced tests we should split it, we saw that already with waterfallgradient, detector and other classes that after growing a bit it makes easy to maintain the code with different files. Creating more 5 files with 30 lines in each one does not sounds a good idea to me ATM.

The reason to put everything in the same cpp was described in my last paraphaph, we can subclass each class, but this makes harder the singletons dependencies that we have. We have some linear startup process dependency. 

I know that right now this single test class is not sustainable and I do not expect to use it until the end, probably I'll split this tests in different cpp files, but all of this files will be functions declarations of the same test class, subclassing is not necessary with this actual test class that I created, as you can see I can access all private and public data without any subclassing.

PR #294, opened 2018-09-14
Update LinkConfiguration, Detector and other things

### LINKED ISSUE #138: Use last successful connection when App starts

patrickelectric:
- Fix #138 
- Correct `ProtocolDetector` implementation
- Correct `LinkConfiguration` debug operator
- Allow `SettingsManager` to work with structures and classes
- Update `LinkConfiguration` to work with `QDataStream`
- Random corrections

### DIFF
--- a/src/detector.cpp
+++ b/src/detector.cpp
@@ -7,6 +7,7 @@
 #include <QSerialPortInfo>
 #include <QUdpSocket>
 
+#include "abstractlinknamespace.h"
 #include "detector.h"
 #include "pingmessage/pingmessage.h"
 #include "pingmessage/pingmessage_ping1D.h"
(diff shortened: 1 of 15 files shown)


src/sensor/ping.h
```diff
@@ -31,6 +30,13 @@ class Ping : public Sensor
      */
     Q_INVOKABLE void connectLink(AbstractLinkNamespace::LinkType connType, const QStringList& connString);
 
+    /**
+     * @brief Add new connection
+     *
+     * @param linkConfig Link Configuration structure
+     */
+    Q_INVOKABLE void connectLink2(LinkConfiguration linkConfig);
```
jaxxzer: Why not just overload the existing function?

src/sensor/ping.cpp
```diff
@@ -72,32 +72,41 @@ Ping::Ping() : Sensor()
...
-            if(_detector.isRunning()) {
-                _detector.exit();
+            if(_detector->isRunning()) {
+                _detector->stop();
             }
         } else {
-            if(!_detector.isRunning()) {
-                _detector.start();
+            if(!_detector->isRunning()) {
+                _detector->scan();
```
jaxxzer: Does this block?

src/sensor/sensor.cpp
```diff
@@ -25,16 +29,30 @@ Sensor::Sensor() :
         this->_connected = false;
         emit this->connectionUpdate();
     });
+
+    auto config = SettingsManager::self()->lastLinkConfiguration();
+    qCDebug(PING_PROTOCOL_SENSOR) << "Loading last configuration connection from settings:" << config;
+    _detector->appendConfiguration(config);
```
jaxxzer: The configuration we load here should be the first one that the detector checks (instant connection at startup). I suggest a public bool Detector::checkLink(LinkConfig& link) that will check a single link. Sensor can call this before calling Detctor::scan.

src/sensor/ping.h
jaxxzer: Nevermind, I see now.

src/sensor/ping.cpp
jaxxzer: don't think so!

src/detector.cpp
```diff
@@ -21,123 +22,162 @@ const QStringList ProtocolDetector::_invalidSerialPortNames(
...
+    port.close();
 
-        // Scan all available ports
-        for (auto& portInfo : portsInfo) {
-            // Do not run with invalid ports
-            if(!isValidPort(portInfo)) {
-                continue;
-            }
+    // Patrick
+    //TODO: move this inside of the big loop
```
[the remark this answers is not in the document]
patrickelectric: lol, I'm one step ahead from myself.

src/sensor/sensor.cpp
patrickelectric: Done.

src/sensor/ping.cpp
```diff
@@ -72,29 +71,26 @@ Ping::Ping() : Sensor()
 
     //connectLink(LinkType::Serial, {"/dev/ttyUSB2", "115200"});
 
-    connect(&_detector, &ProtocolDetector::connectionDetected, this, &Ping::connectLink);
-    _detector.start();
+    connect(_detector, &ProtocolDetector::connectionDetected, this, [this] {
```
jaxxzer: where is connectLink() called after the detector finishes?

jaxxzer:
Here, the detector output continues if the application is closed before connection.

Also, when the application is closed after connection, an exception is thrown:

```
[13:33:07.191] ping.protocol.ping[Debug]: ping.cpp(389) Requesting: 1300
[13:33:07.309] default[Warning]: Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt.
You must not let any exception whatsoever propagate through Qt code.
If that is not possible, in Qt 5 you must at least reimplement
QCoreApplication::notify() and catch all exceptions there.
```

src/sensor/ping.cpp
patrickelectric: the connection is done by sensor class, Ping (as a device class) should not deal with such things, only with things that are for ping device. Besides that, the protocol detection function is something that all devices will probably have, that's why I move it to sensor.

patrickelectric:
@jaxxzer I forgot to disable the main loop of the detection thread. Everything is working fine now.

jaxxzer:
- [ ] remove old files
- [ ] fix manual eth0 configuration
- [ ] put all of our stanzas in one file
- [ ] add mapping config
- [ ] add mapping script

src/settings/settingsmanagerhelper.h
```diff
@@ -34,7 +34,7 @@ public: \
     void NAME(TYPE value) { \
         if(_ ## NAME == value) { return; }\
         _ ## NAME = value; \
-        _settings.setValue(QStringLiteral(#NAME), value); \
+        _settings.setValue(QStringLiteral(#NAME), qVariantFromValue(value)); \
```
jaxxzer: update the comments above.

src/settings/settingsmanagerhelper.h
```diff
@@ -5,14 +5,20 @@
...
  *    Q_PROPERTY(myType myName READ myName WRITE myName NOTIFY myNameChanged)
  *public:
  *    // Get property
  *    myType myName() {
  *        // Always check settings content, make sure that everything is updated
- *        _myName = _settings.value(QStringLiteral("myName")).value<myType>();
- *        return _myName;
+ *        QVariant variant = _settings.value(QStringLiteral(myName)); \
+ *        if(variant.isValid()) { \
+ *            _myName = variant.value<TYPE>(); \
```
jaxxzer: TYPE -> myType

src/link/linkconfiguration.cpp
```diff
@@ -66,9 +66,38 @@ LinkConfiguration::Error LinkConfiguration::error() const
...
-    return text.arg(name(), QString::number(type()), args()->join(":"));
-}
\ No newline at end of file
+    d << text.arg(other.name(), QString::number(other.type()), other.args()->join(":"));;
+    return d;
+}
+
+QDataStream& operator<<(QDataStream &out, const LinkConfiguration linkConfiguration)
+{
+    out << linkConfiguration.configurationStructPtr()->name;
```
jaxxzer: if we didn't use ptr() would it make a copy?

src/link/linkconfiguration.h
```diff
@@ -75,6 +75,12 @@ class LinkConfiguration : public QObject
         return text.arg(name(), QString::number(type()), args()->join(":"));
     }
 
+    QString serialPort() { return (_linkConf.args.size() ? _linkConf.args[0] : QString() ); }
```
jaxxzer: we should validate with LinkConfiguration::type()?

src/link/linkconfiguration.h
```diff
@@ -82,4 +82,26 @@ class LinkConfiguration : public QObject
...
+        const int newId = qRegisterMetaType<LinkConfiguration>("LinkConfiguration");
+        metatypeId.storeRelease(newId);
+        qRegisterMetaTypeStreamOperators<LinkConfiguration>("LinkConfiguration");
+    }
+};
+
+namespace
+{
+LinkConfigurationRegisterStruct _linkConfigurationRegisterStruct;
+}
```
jaxxzer: newline at end of the file.

src/link/linkconfiguration.h
```diff
@@ -82,4 +82,26 @@ class LinkConfiguration : public QObject
...
 QDebug operator<<(QDebug d, const LinkConfiguration& other);
 QDataStream& operator<<(QDataStream &out, const LinkConfiguration linkConfiguration);
-QDataStream& operator>>(QDataStream &in, LinkConfiguration &linkConfiguration);
\ No newline at end of file
+QDataStream& operator>>(QDataStream &in, LinkConfiguration &linkConfiguration);
+
+Q_DECLARE_METATYPE(LinkConfiguration)
+
+// This allows us to register LinkConfiguration before singletons and main function
+struct LinkConfigurationRegisterStruct {
```
jaxxzer: Can you please give me a note what this does and why?

src/link/linkconfiguration.cpp
patrickelectric: Yes, because it'll return a copy.

src/link/linkconfiguration.h
```diff
@@ -52,22 +55,53 @@ class LinkConfiguration : public QObject
...
+};
+
+bool operator==(const LinkConfiguration& first, const LinkConfiguration& second);
+QDebug operator<<(QDebug d, const LinkConfiguration& other);
+QDataStream& operator<<(QDataStream &out, const LinkConfiguration linkConfiguration);
+QDataStream& operator>>(QDataStream &in, LinkConfiguration &linkConfiguration);
+
+Q_DECLARE_METATYPE(LinkConfiguration)
+
+// This allows us to register LinkConfiguration before singletons and main function
```
patrickelectric: @jaxxzer Sorry ! I deleted your comment !

I new comment will be added
```cpp
/**
 * @brief This allows us to register LinkConfiguration before singletons and main function
 * This is something like Q_DECLARE_METATYPE but for operators
 * We should create a Q_DECLARE_METATYPE_STREAM_OPERATORS defined function to deal with future cases
 * For any further explanation:
 * http://doc.qt.io/qt-5/qmetatype.html#Q_DECLARE_METATYPE
 * https://code.woboq.org/qt5/qtbase/src/corelib/kernel/qmetatype.h.html#_M/Q_DECLARE_METATYPE
 * https://code.woboq.org/qt5/qtbase/src/corelib/kernel/qmetatype.h.html#_Z32qRegisterMetaTypeStreamOperatorsPKcPT_
 */
```

PR #295, opened 2018-09-18
Some small correction

patrickelectric:
- Does not show weird values in release builds when not connected
- Does not show depthaxis when value is not set

### DIFF
--- a/src/sensor/ping.h
+++ b/src/sensor/ping.h
@@ -409,29 +409,29 @@ private:
      * @brief Sensor variables
      */
 ///@{
-    QString _ascii_text;
-    QString _nack_msg;
+    QString _ascii_text = QString();
+    QString _nack_msg = QString();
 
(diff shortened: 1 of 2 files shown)


src/sensor/ping.h
```diff
@@ -409,29 +409,29 @@ class Ping : public Sensor
...
+
+    uint8_t _srcId = 0;
+    uint8_t _dstId = 0;
+
+    uint8_t _device_type = 0;
+    uint8_t _device_model = 0;
+    uint16_t _fw_version_major = 0;
+    uint16_t _fw_version_minor = 0;
+
+    uint32_t _distance = 0; // mm
```
jaxxzer: Interesting, I would expect this to be a non-functional change at this point?

patrickelectric: This is only to avoid weird values in the graphic interface when not connected

PR #304, opened 2018-10-03
Fix#123

patrickelectric:


patrickelectric:
This is in WIP, does not represent the final version of this approach.
The VariantTree is what can be reviewed.

patrickelectric:
This PR is close now to the final idea.

src/sensor/ping.cpp
```diff
@@ -86,6 +98,33 @@ Ping::Ping() : Sensor()
     }
 }
 
+void Ping::checkPingConfigurationSettings()
+{
+    for(const auto& key : _pingConfiguration.keys()) {
+        _pingConfiguration[key].value = _pingConfiguration[key].defaultValue;
+    }
+
+    QVariant pingConfigurationVariant = SettingsManager::self()->getMapValue({"Ping", "PingConfiguration", "0"});
```
jaxxzer: Maybe use device id for the path?

src/sensor/ping.h
```diff
@@ -483,8 +485,71 @@ class Ping : public Sensor
     void firmwareUpdatePercentage();
     void flash(const QString& portLocation, const QString& firmwareFile, int baud = 57600, bool verify = true);
 
+    void checkPingConfigurationSettings();
```
jaxxzer: I would reword these to load, save, and apply.

PR #315, opened 2018-10-11
Notification widget

patrickelectric:
Should be done:
- [x] Create a notification manager with everything needed to do so.
- [x] Create a singleton for icons, display information, style, colors.
  - This will helps both cpp and qml to access this kind of information without boilerplate. 
- [x] Move everything related to html requests to C++ as a service.
  - [x] Create a class to deal with updates and new versions.


src/network/networktool.cpp
```diff
@@ -0,0 +1,181 @@
...
+#include <QUrl>
+
+PING_LOGGING_CATEGORY(NETWORKTOOL, "ping.networktool")
+
+void NetworkTool::checkInterfaceUpdate()
+{
+    // It's only necessary to get the gitUserRepo variable only once
+    static bool firstTime = true;
+    static QString gitUserRepo;
+    if(firstTime) {
```
Williangalvani: can't you just check if gitUserRepo is null instead?

src/network/networktool.cpp
```diff
@@ -0,0 +1,181 @@
...
+            gitUserRepo = regexMatch.capturedTexts()[1];
+        } else {
+            qCWarning(NETWORKTOOL) <<
+                                   "Fail to get github user and repository! It'll not be possible to check for updates";
+            return;
+        }
+        firstTime = false;
+    }
+
+    const QUrl url{QStringLiteral("https://api.github.com/repos/%1/releases").arg(gitUserRepo)};
```
Williangalvani: what happens if the first part failed and it gets here? fails silently?

src/notification/notificationmodel.cpp
```diff
@@ -0,0 +1,114 @@
...
+bool NotificationModel::insertRows(int row, int count, const QModelIndex& parent)
+{
+
+    // Check for valid input
+    if (count < 1 || row < 0 || row > rowCount(parent)) {
+        return false;
+    }
+
+    // Add all request rows
+    while(count --> 0) {
```
Williangalvani:
```suggestion
    while(count-- > 0) {
```

qml/DisplaySettings.qml
```diff
@@ -65,21 +67,15 @@ Item {
...
                         Layout.fillWidth: true
                         model: ["Dark", "Light"]
-                        onCurrentTextChanged: {
-                            if (currentIndex) {
-                                Style.useLightStyle()
-                            } else {
-                                Style.useDarkStyle()
-                            }
-                        }
+                        onCurrentTextChanged: StyleManager.isDark = !currentIndex
```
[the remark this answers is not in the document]
patrickelectric: Why this is wrong ?

qml/PingNotificationArea.qml
```diff
@@ -0,0 +1,174 @@
...
+            ScrollView {
+                id: scroll
+                clip: true
+                Layout.maximumHeight: window.height/2
+
+                ScrollBar.vertical: ScrollBar {
+                    // This will help to find which ScrollBar is not being used
+                    objectName: "verticalBar"
+                    interactive: false
+                    policy: ScrollBar.AlwaysOn
```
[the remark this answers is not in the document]
patrickelectric: This helps the user to see if it exists new notifications without scrolling.

qml/PingNotificationArea.qml
```diff
@@ -0,0 +1,174 @@
...
+                }
+
+                Component.onCompleted: {
+                    // This will delete all scrollbars that are not used
+                    // Doing: ScrollBar.vertical: ScrollBar {
+                    // does not remove old childrens
+                    // This is possible to check with gammaray
+                    // where it's possible to see the QQuickScrollBar items
+                    // that overlap right right and bottom part of the widget
+                    for(var i in scroll.data) {
```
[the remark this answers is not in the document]
patrickelectric: This syntax does not exist in qml:
```qml
import QtQuick 2.7

Item {
    id: root

    Component.onCompleted: {
        var a = [3, 2, 1]
        for(var i of a) {
            print(i)
        }
    }
}

```

qml/PingNotificationArea.qml
```diff
@@ -0,0 +1,174 @@
...
+                                    }
+
+
+                                    MouseArea {
+                                        anchors.fill: parent
+                                        // Do not handle clicks to expand text is a link
+                                        acceptedButtons: parent.hoveredLink ? Qt.NoButton : Qt.LeftButton
+                                        cursorShape: parent.hoveredLink ? Qt.PointingHandCursor : Qt.ArrowCursor
+                                        onClicked: {
+                                            if(!textItem.isBigText) {
```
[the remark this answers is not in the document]
patrickelectric: This will make the indentation of the block bigger, and as consequence worst if the following lines get more complex.

src/network/networktool.cpp
patrickelectric: What part ? Are you talking about `QStringLiteral` ?

src/notification/notificationmodel.cpp
patrickelectric: I call it "big arrow operator".

patrickelectric:
@Williangalvani 	corrections and comments done.

src/network/networktool.cpp
Williangalvani: it still makes the request even with gitUserRepo empty?
nevermind, didn't notice the return there.

qml/DisplaySettings.qml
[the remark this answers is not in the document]
Williangalvani: well, first because you are processing the event as a single 'toggling',
then you are using boolean operators to toggle it between 1 and 0 ...

wouldn't work if you decided to add, for instance, a third colorblind mode
I don't think it is going to cause any trouble, though.

src/network/networktool.cpp
Williangalvani: I meant it could try to make requests without gitUserRepo set. looks fine now.

qml/PingNotificationArea.qml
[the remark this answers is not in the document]
Williangalvani: don't new ones show up on the top, pushing the old ones down?

src/notification/notificationmodel.cpp
Williangalvani: nice

qml/PingNotificationArea.qml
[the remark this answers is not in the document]
patrickelectric: No, also this will make harder to read a selected event.

qml/PingNotificationArea.qml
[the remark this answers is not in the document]
Williangalvani: sad

qml/Ping1DVisualizer.qml
```diff
@@ -52,7 +53,7 @@ Item {
                 transform: Rotation { origin.x: height/2; angle: 90}
                 gradient: Gradient {
                     GradientStop { position: 0.3; color: "transparent" }
-                    GradientStop { position: 0.5; color: Style.primaryColor } // Not working with material
+                    GradientStop { position: 0.5; color: StyleManager.secondaryColor } // Not working with material
```
jaxxzer: is the comment still relevant?

jaxxzer:
Running doxygen gives `error: Project logo '../qml/imgs/ping.svg' specified by PROJECT_LOGO does not exist!`

qml/Ping1DVisualizer.qml
patrickelectric: Yep.

patrickelectric:
@jaxxzer This PR do not changes ping.svg path, file or the Doxyfile. Besides that, try to run Doxygen inside the Doxyfile.

src/network/networkmanager.h
```diff
@@ -0,0 +1,48 @@
...
+     */
+    static NetworkManager* self();
+
+    /**
+     * @brief Do json requests
+     * @param url URL to be requested
+     * @param function that will be called
+     *
+     */
+    static void requestJson(const QUrl& url, std::function<void(QJsonDocument&)> function);
```
jaxxzer: `function` -> `callback`?

qml/main.qml
```diff
@@ -6,14 +6,18 @@ import QtGraphicalEffects 1.0
...
 ApplicationWindow {
     id: window
+    objectName: "window"
     title: "Ping Interface"
     visible: true
     height: 600
     width: 800
 
-    Material.theme: Style.theme
+    Material.theme: StyleManager.theme
```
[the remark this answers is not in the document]
patrickelectric: Material is already embedded in each controller, redoing that will be necessary to create each controller for our application just to not use Material.
 - https://github.com/qt/qtquickcontrols2/blob/5.11/src/imports/controls/material/material.pri

src/network/networkmanager.h
patrickelectric: Yes, that's the idea. I'm using the function name because this variable is a std::function. A `callback`variable can be many things and a function is a function.

qml/Ping1DVisualizer.qml
patrickelectric: Because they are not working with Material. As the comment says.
Probably a bug in the signal propagation of Material, I'm not sure.

jaxxzer:
Generally looks ok on first pass.
I am not sure about the NetworkManager and NetworkTool names, separation of roles and interfaces, but we can mature them more as needed.

qml/main.qml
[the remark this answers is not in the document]
jaxxzer: ok!

src/network/networktool.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include <QUrl>
+
+PING_LOGGING_CATEGORY(NETWORKTOOL, "ping.networktool")
+
+void NetworkTool::checkInterfaceUpdate()
+{
+    static QString gitUserRepo;
+    if(gitUserRepo.isEmpty()) {
+        //https://github.com/user/repo.git
+        const static QRegularExpression regex("github.com/(.*).git");
```
jaxxzer:
```suggestion
        const static QRegularExpression regex("github.com/(.*)");
```
I don't put `.git`, it's not needed here or there.

src/network/networktool.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include <QTimer>
+#include <QUrl>
+
+PING_LOGGING_CATEGORY(NETWORKTOOL, "ping.networktool")
+
+void NetworkTool::checkInterfaceUpdate()
+{
+    static QString gitUserRepo;
+    if(gitUserRepo.isEmpty()) {
+        //https://github.com/user/repo.git
```
jaxxzer:
```suggestion
        //https://github.com/user/repo
```

src/network/networktool.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include <QJsonDocument>
+#include <QRegularExpression>
+#include <QTimer>
+#include <QUrl>
+
+PING_LOGGING_CATEGORY(NETWORKTOOL, "ping.networktool")
+
+void NetworkTool::checkInterfaceUpdate()
+{
+    static QString gitUserRepo;
```
jaxxzer:
```suggestion
    // check if the remote named `origin` is a github repository (determined at qmake-time) so we may use
    // the github api to check for available versions
    static QString gitUserRepo;
```

src/network/networktool.cpp
```diff
@@ -0,0 +1,178 @@
...
+    );
+}
+
+void NetworkTool::checkNewVersionInGitHubPayload(QJsonDocument& jsonDocument)
+{
+    const static QString projectTag = QStringLiteral(GIT_TAG);
+    static QRegularExpression releaseTagRegex;
+    // Check if version start with 't' from test version
+    // For everything we can presume normal release
+    if(projectTag.contains('t') == 0) {
```
jaxxzer: // Check if version starts with 't' for test version
// Check if version starts with 'v' for release version

src/network/networktool.cpp
```diff
@@ -0,0 +1,178 @@
...
+        QJsonValue json;
+    } lastReleaseAvailable;
+    lastReleaseAvailable.version = actualVersion;
+
+    // Check payload
+    for(const QJsonValue& versionPayload : versionsAvailable) {
+        auto versionString = versionPayload[QStringLiteral("tag_name")].toString();
+        qCDebug(NETWORKTOOL) << "Testing version:" << versionString;
+
+        regexMatch = releaseTagRegex.match(versionString);
```
jaxxzer: If we are currently running a test release, we will only be checking for test releases upstream
If we are currently running a regular release, we will only be checking for regular releases upstream (see above lines 46 and 49). Is this ok?

src/network/networktool.cpp
```diff
@@ -0,0 +1,178 @@
...
+        if(!regexMatch.hasMatch()) {
+            continue;
+        }
+        qCDebug(NETWORKTOOL) << "Possible new version.";
+
+        auto version = versionString.right(1).toFloat();
```
jaxxzer: I think we need to talk about our versioning standards organization-wide. I like the simplicity of being able to simply parse a float, but I have a preference for Major.Minor.Patch semantic versioning.
https://semver.org/

src/network/networktool.cpp
patrickelectric: Yes, that's the idea. Anyone that want to be bleeding edge or want to be a beta testerof ping-viewer will receive notifications about new test releases (a.k.a Beta).

src/network/networktool.cpp
patrickelectric: In my mind, it depends of what kind of release cycle we want to use in the project.
For APIs, semver makes sense. For HUGE projects cmsver. And for simple projects calver or this one that I'm using, that is just new_versions.patches;

src/network/networktool.cpp
jaxxzer: the regex will capture both cases with my suggested edit.

src/network/networktool.cpp
jaxxzer: But these same people will not catch the stable release, even when it is newer than their beta/test.

src/network/networktool.cpp
jaxxzer: @Williangalvani what is your thought?

src/notification/notificationmodel.cpp
jaxxzer: Do you seriously like this @patrickelectric? The whitespace here looks insane to me. I'm surprised it is even legal.

src/notification/notificationmodel.h
```diff
@@ -0,0 +1,133 @@
...
+    Q_INVOKABLE void remove(int index);
+
+    /**
+     * @brief Remove all items
+     *
+     */
+    Q_INVOKABLE void removeAll();
+
+private:
+    Q_DISABLE_COPY(NotificationModel)
```
jaxxzer: Why is this necessary?

jaxxzer: http://doc.qt.io/qt-5/qobject.html#Q_DISABLE_COPY

I see now, but why don't we do this on NotificationManager or our other QObject derived classes?

src/notification/notificationmodel.cpp
```diff
@@ -0,0 +1,114 @@
...
+        for(auto* vector : _rows) {
```
jaxxzer: why do we use `auto*` here instead of just `auto`?

src/network/networktool.cpp
patrickelectric: The problem is that "https://api.github.com/repos/bluerobotics/ping-viewer.git/releases" does not exist.

src/notification/notificationmodel.cpp
patrickelectric: To not get a copy of it, AFAIK the compiler will set auto to T.

src/notification/notificationmodel.h
patrickelectric: NotificationManager is a singleton, it's not possible to copy by it definition.
But about the other QObjects that we have yes. This line is a result based on Qt source code study of models that already exist.

src/network/networktool.cpp
patrickelectric: Done.

patrickelectric:
@jaxxzer Check if you agree with the corrections, let me know if something else need to be changed or merge as will.

src/notification/notificationmodel.h
jaxxzer: The document that I linked says that any QObject should never be copied, not just singletons.

src/notification/notificationmodel.cpp
jaxxzer: That's cool.

src/notification/notificationmodel.h
patrickelectric: Check this line:
https://github.com/bluerobotics/ping-viewer/pull/315/files#diff-ef6adb74577178213a0ad3d437b8b25cR45

src/network/networktool.cpp
jaxxzer: @patrickelectric I agree generally, but I think there is benefit to consistency in our organization or something we can add to our guidelines. Something that our users and engineers (current and future) can get used to. I agree that x.x.x may be more than we need here, but it's not _that_ excessive in my opinion and will be able to cover every project we do.

We don't need to reach decision before merge, but before we cut first stable.

ping @Williangalvani.

Williangalvani: I agree with @jaxxzer that consistency is important.
I understand that it is not required in this project right now, and a little more annoying to code with, but I vote for semver.
Or maybe we could have x.x standard for smaller projects?
Also, is your code assuming the version is only 1 digit?

src/notification/notificationmodel.h
jaxxzer: hehe, it looks like this is the same effect and purpose of Q_DISABLE_COPY, so why not use that?

patrickelectric: Good point, I thought that was a different kind of disable copy but they are not. Will change it to make it uniform.

PR #316, opened 2018-10-11
Tools: Add testcomments

patrickelectric:
-Test PRs to have at least 10% of comments in .cpp files

### DIFF
--- a/tools/runtests.sh
+++ b/tools/runtests.sh
@@ -22,6 +22,9 @@ fi
 echob "Running style script:"
 ${scriptpath}/testdoxygen.sh || exit 1
 
+echob "Check comments:"
+${scriptpath}/testcomments.sh || exit 1
+
 echob "Compile code in test mode:"
 build_test="$projectpath/build_test"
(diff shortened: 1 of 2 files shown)


jaxxzer:
oof. I feel like this could be a bad idea.

Why 10%? Why the .cpp in particular? 

The doxygen at least goes in the header. I would prefer a check that everything is doxygenated (we have this already?). 

The cpp we can take on a case-by-case basis and request more comments if necessary in the review. I don't think the algorithm for 'enough comments' can be so simple as x%.

Williangalvani:
What happens if the PR is a simple fix and the patch has 2 lines and no comments?

patrickelectric:
This is just something that I use in my internal projects, ok, closing PR.

jaxxzer:
Do we have doxygen check?

patrickelectric:
Yes, https://github.com/bluerobotics/ping-viewer/blob/master/tools/testdoxygen.sh.

PR #317, opened 2018-10-16
Small corrections

patrickelectric:
- This also allow any previous link configuration to be used by detector

### DIFF
--- a/src/sensor/ping.cpp
+++ b/src/sensor/ping.cpp
@@ -106,9 +106,8 @@ Ping::Ping() : Sensor()
     // Load last successful connection
     auto config = SettingsManager::self()->lastLinkConfiguration();
     qCDebug(PING_PROTOCOL_PING) << "Loading last configuration connection from settings:" << config;
-    if(!detector()->checkLink(config)) {
-        detectorThread()->start();
-    }
+    addDetectionLink(config);
+    detectorThread()->start();
 }
 
 void Ping::loadLastPingConfigurationSettings()
--- a/src/sensor/ping.cpp
+++ b/src/sensor/ping.cpp
@@ -141,6 +141,21 @@ void Ping::updatePingConfigurationSettings()
     }
 }
 
+void Ping::addDetectionLink(const LinkConfiguration& linkConfiguration)
+{
+    if(linkConfiguration.isValid()) {
+        detector()->appendConfiguration(linkConfiguration);
+    } else {
+        qCDebug(PING_PROTOCOL_PING) << "Invalid configuration:" << linkConfiguration.errorToString();
+    }
+}
+
+void Ping::addDetectionLink(LinkType connType, const QStringList& connString)
+{
+    LinkConfiguration linkConfiguration{connType, connString};
+    addDetectionLink(linkConfiguration);
+}
+
 void Ping::connectLink(LinkType connType, const QStringList& connString)
 {
     Sensor::connectLink(LinkConfiguration{connType, connString});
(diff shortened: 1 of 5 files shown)


src/sensor/ping.cpp
```diff
@@ -138,6 +138,21 @@ void Ping::updatePingConfigurationSettings()
     }
 }
 
+void Ping::tryConnectLink(const LinkConfiguration& linkConfiguration)
```
jaxxzer: Think we need a better name for the function.

I suggest addDetectionLink

qml/MainMenu.qml
```diff
@@ -32,7 +32,7 @@ Item {
 
         // Transform arguments to Array and remove connectionTypeEnum from it
         var nextArgs = Array.prototype.slice.call(arguments).slice(1)
-        ping.connectLink(connectionTypeEnum, nextArgs)
+        ping.tryConnectLink(connectionTypeEnum, nextArgs)
```
jaxxzer: I'm not sure if this patch is actually appropriate. I think if the user is entering an explicit connection, autodetect should be disabled.

src/sensor/ping.cpp
```diff
@@ -54,6 +54,7 @@ Ping::Ping() : Sensor()
     //connectLink(LinkType::Serial, {"/dev/ttyUSB2", "115200"});
 
     connect(detector(), &ProtocolDetector::connectionDetected, this, [this] {
+        qCDebug(PING_PROTOCOL_PING) << "New device detected";
```
jaxxzer: Can we print any more information about the connection?

qml/MainMenu.qml
patrickelectric: In noise links, the communication may fail and will not try to detect again, this will make the user experience not clear when adding a URL link that can fail n times.

src/sensor/ping.cpp
```diff
@@ -103,9 +103,8 @@ Ping::Ping() : Sensor()
     // Load last successful connection
     auto config = SettingsManager::self()->lastLinkConfiguration();
     qCDebug(PING_PROTOCOL_PING) << "Loading last configuration connection from settings:" << config;
-    if(!detector()->checkLink(config)) {
-        detectorThread()->start();
-    }
+    tryConnectLink(config);
```
jaxxzer: This is a regression, no?

The configuration here is appended to the list to detect, not guaranteed to be checked first. We want to check this link explicitly first before detection process starts.

src/sensor/ping.cpp
patrickelectric: The connection information, when a new device is connected appears in protocoldetector.

patrickelectric:
@jaxxzer 	can you review this ?

qml/MainMenu.qml
jaxxzer: The noise does not matter, that is outside the scope of this application.

If I want to connect to udp ip:port specifically, and I explicitly specify it, that is what we should do!

If the user enters a udp or serial configuration here, we should not _detect_ on that configuration, we should skip detect, and just _connect_ on that configuration, whether there is noise, garbage, or nothing. The user said to do it, so do it. This makes the user experience clear.

jaxxzer: The implication if the user interacts here is that the user knows where/how the sensor is connected. Therefore we do not need to detect.

patrickelectric: I agree with you, I will do further tests with UDP to take a closer look on it.
Commit removed.

PR #324, opened 2018-11-05
Fix #323

### LINKED ISSUE #323: Fix memory leak from logger widget

patrickelectric:
- Change LogListModel to use QAbstractListModel
- Add filter function
- This method also decreases the cpu usage in ~10%.


jaxxzer:
nice!

PR #325, opened 2018-11-05
MainMenu: Hide serial or UDP settings when necessary

patrickelectric:

### DIFF
--- a/qml/MainMenu.qml
+++ b/qml/MainMenu.qml
@@ -303,6 +303,7 @@ Item {
 
                 RowLayout {
                     id: serialLayout
+                    visible: serialLayout.enabled
                     spacing: 5
                     Layout.columnSpan:  5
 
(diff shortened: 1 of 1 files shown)


jaxxzer:
What is the idea here?
When should serial/udp be visible and when should they not be?

patrickelectric:
This PR turn udp visibility off when the connection is Serial and vice-versa.

jaxxzer:
What if the user wants to switch the active connection?

If the user can change to a different serial configuration while connected to serial.. why must we not allow them to also change to a network connection?

The question might be, should we allow them to change the connection at all when a connection exists?

patrickelectric:
>What if the user wants to switch the active connection?

Use the combobox to select the connection type.

> If the user can change to a different serial configuration while connected to serial.. why must we not allow them to also change to a network connection?

What ? They can change to udp or serial, and configure that is enabled.

> The question might be, should we allow them to change the connection at all when a connection exists?

Well, this is not what this PR is about, it just to remove unnecessary stuff when a connection configuration is enabled. And, besides that, I don't see why not.

jaxxzer:
I was confused about what this was doing.

PR #326, opened 2018-11-06
Fix #318

### LINKED ISSUE #318: Check for nacks and lost messages

patrickelectric:


src/sensor/ping.h
```diff
@@ -552,4 +566,18 @@ class Ping : public Sensor
...
+
+    // total of lost messages
+    int _lostMessages = 0;
+
+    struct messageStatus {
+        // Requested and acknowledge
+        int ack = 0;
+        // Requested and not acknowledge
+        int nack = 0;
+        // Number of requested messages
```
jaxxzer: Number of requested messages with no response.

patrickelectric: Not necessary, this is the number of messages waiting to be received after the request. It's not possible to confirm that we will not receive and they are lost. The only case to confirm such comment, is to have a vector of timestamps of requests and use a time threshold to confirm the "lost message" state.

jaxxzer: Ok agree. Maybe 'with outstanding replies', 'waiting replies' or maybe similar. My argument/suggestion here is that the current comment is not correct, we have lines with requested messages --. The current comment indicates that the number will grow with each request, not shrink.

patrickelectric: Variable moved to `waiting` and comment to: `// Number of waiting replies`

PR #333, opened 2018-11-08
Use semver and similar types to identify version

patrickelectric:


src/network/networktool.cpp
```diff
@@ -38,19 +38,38 @@ void NetworkTool::checkNewVersionInGitHubPayload(QJsonDocument& jsonDocument)
...
     // Check if version is a normal release
     // For everything else we can presume test release
-    static QRegularExpression releaseTagRegex(QStringLiteral(R"([v]\d+\.\d+)"));
+    // This regex will check for vX, vX.X and vX.X.X
+    static QRegularExpression releaseTagRegex(QStringLiteral(R"([v]\d+(\.\d+){0,})"));
     QRegularExpressionMatch regexMatch = releaseTagRegex.match(projectTag);
     // Check if this actual version is a real release (normal or test)
     // Alarm people to download the correction version !
-    float actualVersion = regexMatch.hasMatch() ? projectTag.right(1).toFloat() : -1.0;
+    int actualVersion = regexMatch.hasMatch() ? semverToInt(projectTag) : -1.0;
```
jaxxzer: You need to change also https://github.com/bluerobotics/ping-viewer/pull/333/commits/e4acb8c3ec5277f62f7e84bb31e9ce13d27f3399#diff-55288e377d501245cba94fbcda302a67R102

src/network/networktool.cpp
```diff
@@ -38,19 +38,38 @@ void NetworkTool::checkNewVersionInGitHubPayload(QJsonDocument& jsonDocument)
...
+        {
+            result += subVersion[i].toInt()<<(3-i);
+        }
+        return result;
+    };
+
     // Check if version is a normal release
     // For everything else we can presume test release
-    static QRegularExpression releaseTagRegex(QStringLiteral(R"([v]\d+\.\d+)"));
+    // This regex will check for vX, vX.X and vX.X.X
```
jaxxzer: This <s>does not seem necessary at all</s> seems wrong to check for anything but vx.x.x.

When the goal is to make a consistent standard, it does not make sense to compare version x with x.x.x. We will never do it. It's like unused code.

src/network/networktool.cpp
```diff
@@ -38,19 +38,38 @@ void NetworkTool::checkNewVersionInGitHubPayload(QJsonDocument& jsonDocument)
...
+        // Check if subVersion is valid
+        if(subVersion.length() > 3)
+        {
+            return result;
+        }
+
+        // Transform version to int
+        for(int i = 0; i < subVersion.length(); i++)
+        {
+            result += subVersion[i].toInt()<<(3-i);
```
jaxxzer: This is not right at all. How did you come up with this?

version 1.2.3 is equivalent to 2.1.1

version 1.11.0 is greater than version 2.0.0

jaxxzer: Can we do something analogous to this?
https://github.com/mavlink/qgroundcontrol/pull/6913/files#diff-ae5366dfccfee1039895edb6b2131a6eR731

PR #339, opened 2018-11-14
Fix loading style from stylemanager

patrickelectric:


src/style/stylemanager.cpp
```diff
@@ -11,7 +11,7 @@ PING_LOGGING_CATEGORY(STYLEMANAGER, "ping.stylemanager")
 
 StyleManager::StyleManager()
 {
-    connect(this, &StyleManager::isDarkChanged, this, [&] {
+    connect(this, &StyleManager::isDarkChanged, this, [this] {
```
jaxxzer: I read [this](https://en.cppreference.com/w/cpp/language/lambda#Lambda_capture) as the opposite of what you have done/noted here.

`this` will capture the entire object

`&` will implicitly capture only what is used (which in this case is only the object so as-is the patch doesn't do anything different).

patrickelectric: `[&, this]` is equal to `&`, `&` will get everything inside the scope, `this` will only capture the pointer of the object. We not getting variables inside the scope but only the functions and variables of the class.

jaxxzer: > & will get everything inside the scope

Inside of what scope?

I read the cppreference as inside the scope of the lambda function itself, not the scope where the lambda is defined. Thus, `&` will get you only the variables that you need, which is what you desire.

jaxxzer:
```
struct S2 { void f(int i); };
void S2::f(int i)
{
    [&]{};          // OK: by-reference capture default
    [&, i]{};       // OK: by-reference capture, except i is captured by copy
    [&, &i] {};     // Error: by-reference capture when by-reference is the default
    [&, this] {};   // OK, equivalent to [&]
    [&, this, i]{}; // OK, equivalent to [&, i]
}
```

I think this example verifies my point. (look at the last one)

patrickelectric: Also: https://stackoverflow.com/questions/33575563/c-lambda-capture-this-vs-capture-by-reference/33576211

PR #347, opened 2018-11-28
ping: Save last link configuration for manual input

### LINKED ISSUE #311: Last link configuration is not working for udp
1. Connects with UDP.
2. Open the interface again.
3. Will not connect.

patrickelectric:
Fix #311

### DIFF
--- a/src/sensor/ping.cpp
+++ b/src/sensor/ping.cpp
@@ -67,25 +67,7 @@ Ping::Ping() : Sensor()
 
     //connectLink(LinkType::Serial, {"/dev/ttyUSB2", "115200"});
 
-    connect(detector(), &ProtocolDetector::connectionDetected, this, [this] {
-        qCDebug(PING_PROTOCOL_PING) << "New device detected";
-        setAutoDetect(false);
-
-        // Request device information
-        request(Ping1DNamespace::Mode_auto);
-        request(Ping1DNamespace::Profile);
-        request(Ping1DNamespace::Firmware_version);
-        request(Ping1DNamespace::Device_id);
-        request(Ping1DNamespace::Speed_of_sound);
-
-        if(link()->isWritable())
-        {
-            // Save configuration
-            SettingsManager::self()->lastLinkConfiguration(*link()->configuration());
-            // Start periodic request timer
-            _periodicRequestTimer.start();
-        }
-    });
+    connect(this, &Sensor::connectionOpen, this, &Ping::startPreConfigurationProcess);
 
     // Wait for device id to load the correct settings
     connect(this, &Ping::srcIdUpdate, this, &Ping::setLastPingConfiguration);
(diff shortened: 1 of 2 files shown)


src/sensor/ping.cpp
```diff
@@ -167,6 +167,7 @@ void Ping::addDetectionLink(LinkType connType, const QStringList& connString)
 void Ping::connectLink(LinkType connType, const QStringList& connString)
 {
     Sensor::connectLink(LinkConfiguration{connType, connString});
+    SettingsManager::self()->lastLinkConfiguration(*link()->configuration());
```
jaxxzer: I believe we should connect connectLink to connectionDetected, and move the first requests here.

patrickelectric: This is already done. This function is called by the QML part, because sensor does not provide it as Q_INVOKABLE. This is a result of a bug where QML can't handle functions with default values or with the same name.

jaxxzer:
I am concerned about the direction of this 'connect sensor' process and the structure around it.

I feel like the connection and logic should be **simply** `ProtocolDetector::connectionDetected -> Ping::connectLink`.

In this patch we have `ProtocolDetector::connectionDetected -> **Sensor**::connectLink` and also `ProtocolDetector::connectionDetected -> Ping::startPreconfigurationProcess`

We have a third overload of `connectLink` in Ping that does not match/exist in the parent Sensor class.

I have two questions:
- Is there any guarantee of the order that the connects run (in Sensor and Ping, both are connecting independently) after ProtocolDetector::connectionDetected is signaled?
- Do you think we can unify the handling of this between sensor and Ping? Can we make Ping override connectLink, so that Ping's method is called in the `ProtocolDetector::connectionDetected -> **Sensor**::connectLink` [connection](https://github.com/bluerobotics/ping-viewer/blob/8ef71de26cc8d83dd88c9573212c565fd037e5df/src/sensor/sensor.cpp#L34)?

src/sensor/ping.cpp
```diff
@@ -120,6 +102,34 @@ Ping::Ping() : Sensor()
     detectorThread()->start();
 }
 
+void Ping::startPreConfigurationProcess()
+{
+    qCDebug(PING_PROTOCOL_PING) << "Start pre configuration task and requests.";
+    if(!link()->isWritable()) {
+        qCDebug(PING_PROTOCOL_PING) << "It's not possible to set last configuration when link is writable.";
```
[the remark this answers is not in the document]
patrickelectric: @jaxxzer 	done.

patrickelectric:
> I feel like the connection and logic should be simply ProtocolDetector::connectionDetected -> Ping::connectLink.

This is done by the sensor part, because all sensors will need connections and protocol detections AFAIK. Moving this to ping will make all sensors definition to have the same code, that could be in sensor base class.

> In this patch we have two slots that are called ProtocolDetector::connectionDetected -> **Sensor**::connectLink and also ProtocolDetector::connectionDetected -> Ping::startPreconfigurationProcess

This was done based on the code that existed. Moving it to use sensor now.

> We have a third overload of connectLink in Ping that does not match/exist in the parent Sensor class, and is also surprisingly not involved during the autodetect/connection process.

This was because of some problems with QML and c++ structures, I can do some work to fix that.

src/sensor/ping.cpp
```diff
@@ -120,6 +99,34 @@ Ping::Ping() : Sensor()
...
+    request(Ping1DNamespace::Speed_of_sound);
+
+    // Start periodic request timer
+    _periodicRequestTimer.start();
+
+    // Save configuration
+    SettingsManager::self()->lastLinkConfiguration(*link()->configuration());
+
+    // Start configuration
+    setLastPingConfiguration();
```
jaxxzer: Are we guaranteeded that we will have the correct srcId for the device at this point so that we load the correct settings?

patrickelectric:
@jaxxzer thank you for making the PR awesome and figuring out so much things !

PR #348, opened 2018-11-28
ping: Do not request or save configuration if running a replay

### LINKED ISSUE #343: Application tries to configure file link as a normal sensor

patrickelectric:
Fix #343

### DIFF
--- a/src/sensor/ping.cpp
+++ b/src/sensor/ping.cpp
@@ -70,7 +70,6 @@ Ping::Ping() : Sensor()
     connect(detector(), &ProtocolDetector::connectionDetected, this, [this] {
         qCDebug(PING_PROTOCOL_PING) << "New device detected";
         setAutoDetect(false);
-        SettingsManager::self()->lastLinkConfiguration(*link()->configuration());
 
         // Request device information
         request(Ping1DNamespace::Mode_auto);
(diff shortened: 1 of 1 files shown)


patrickelectric:
We do not have write steps, also, doing this, this will not make all configuration timers to stop, resulting in a series of warning messages.

jaxxzer:
setLastPingConfiguration is the write step. If the link is not writable, it should be handled inside of that function. You may also handle !iswritable() elsewhere, but it should be at least handled in setLastPingConfiguration.

patrickelectric:
@jaxxzer 	did you check the last commit ?

jaxxzer:
Not before I replied. Will check

src/sensor/ping.cpp
```diff
@@ -487,6 +472,29 @@ void Ping::request(int id)
 
 void Ping::setLastPingConfiguration()
 {
+    if(!link()->isWritable()) {
```
jaxxzer:
```suggestion
    static bool once = false;
    if (once) {
        return; // or *disconnect*
    }
    
    if(!link()->isWritable()) {
```

jaxxzer: This will be called on every single message rxed from the device.

jaxxzer: your 'once' did not encapsulate the entire code that needed to be run 'once'. We were trying to set device values from saved settings every time a message was rxed.

patrickelectric:
Fixed.

src/sensor/ping.cpp
```diff
@@ -504,6 +489,35 @@ void Ping::request(int id)
...
+    lastSrcId = _srcId;
+
+    if(!link()->isWritable()) {
+        qCDebug(PING_PROTOCOL_PING) << "It's not possible to set last configuration when link is writable.";
+        return;
+    }
+
+    // Load last configuration
+    static bool once = false;
+    if(!once) {
```
jaxxzer: I think this condition should not be here anymore with the lastSrcId check above.

patrickelectric: I did it, my I have done some ctrl+z and this got lost.

PR #350, opened 2018-11-29
Check if a new firmware version is available

patrickelectric:


src/sensor/ping.cpp
```diff
@@ -564,6 +581,32 @@ void Ping::writeMessage(const PingMessage &msg)
...
+        QString downloadUrl;
+    } versionAvailable;
+
+    auto filesPayload = jsonDocument.array();
+    for(const QJsonValue& filePayload : filesPayload) {
+        qCDebug(PING_PROTOCOL_PING) << filePayload["name"].toString();
+
+        // Get version from Ping_V(major).(patch)_115kb.hex where (major).(patch) is <version>
+        static const QRegularExpression versionRegex(QStringLiteral(R"(Ping_V(?<version>\d+\.\d+)_115kb\.hex)"));
+        versionAvailable.version = versionRegex.match(filePayload["name"].toString()).captured("version").toFloat();
```
jaxxzer: There is no guarantee that the json will be ordered in ascending version. You are making that assumption here, so you could just take array.last() instead of looping through each. (but let's stop making that assumption):
```suggestion
        versionAvailable = versionRegex.match(filePayload["name"].toString()).captured("version").toFloat();
        if (versionAvaliable > latestVersionAvailable.version) {
            latestVersionAvailable.version = versionAvailable;
            latestVersionAvailable.downloadUrl = ...
        }
```

patrickelectric: True, probably I lost some code after moving things around, this was done in the initial version. Going to fix it soon.

patrickelectric:
I agree with the auto update handled by the graphic interface. This is a initial version indeed.

PR #361, opened 2018-12-17
Add auto flash feature

patrickelectric:
image

image
image

image

### DIFF
--- a/qml/FirmwareUpdate.qml
+++ b/qml/FirmwareUpdate.qml
@@ -7,6 +7,7 @@ import QtQuick.Dialogs 1.2
 import QtQuick.Layouts 1.3
 import Qt.labs.settings 1.0
 
+import Flasher 1.0
 import SettingsManager 1.0
 
 Item {
@@ -14,17 +15,31 @@ Item {
     height: settingsLayout.height
     width: settingsLayout.width
     property var ping: null
+    property var running: false
 
     Connections {
-        target: ping
+        target: ping.flasher
         onFlashProgress: {
-            flashProgress.indeterminate = false
+            running = true
             flashProgress.value = progress
         }
-        onFlashComplete: {
-            console.log("flash complete!")
-            flashProgress.value = 0.0
-            flashProgress.indeterminate = true
+        onFlashStateChanged: {
+            switch(state) {
+                case Flasher.Error:
+                    print("Flash procedure failed!")
+                    print("Error: ", ping.flasher.error)
+
+                case Flasher.FlashFinished:
+                    running = false
+                    flashProgress.value = 0.0
+                    break;
+                case Flasher.Flashing:
+                case Flasher.StartingFlash:
+                    running = true
+                    break;
+                default:
+                    running = true
+            }
         }
     }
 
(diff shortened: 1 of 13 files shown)


patrickelectric:
Tested and working on windows.

src/flash/flasher.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    QSerialPortInfo pInfo(_link.serialPort());
+    auto portLocation = pInfo.systemLocation();
+
+    auto baudRate = QString::number(_baudRate);
+    auto verifyArgument = _verify ? QStringLiteral("-v") : QString();
+
+    QFileInfo firmwareFileInfo(_firmwareFilePath);
+    if(!firmwareFileInfo.exists()) {
+        qCCritical(FLASH) << "Firmware file dows not exist:" << _firmwareFilePath;
```
jaxxzer: spelling `dows`

src/flash/flasher.h
```diff
@@ -0,0 +1,95 @@
...
+        Idle,
+        StartingFlash,
+        Flashing,
+        FlashFinished,
+        ErrorFlashing,
+    } states;
+    Q_ENUMS(States)
+
+signals:
+    void downloadProgress(float progress);
```
jaxxzer: unused?

src/flash/flasher.cpp
```diff
@@ -0,0 +1,136 @@
...
+#ifdef Q_OS_WIN
+    return binPath + "/stm32flash.exe";
+#else
+    return binPath + "/stm32flash";
+#endif
+}
+
+void Flasher::flash()
+{
+    emit flashStateUpdate(Flasher::StartingFlash);
```
jaxxzer: `Flasher::` is unnecessary here, you are already in the namespace.

src/flash/flasher.h
```diff
@@ -0,0 +1,95 @@
...
+     *
+     * @param verify
+     */
+    void setVerify(bool verify);
+
+    /**
+     * @brief Defines flash state
+     *
+     */
+    enum States {
```
jaxxzer: It looks like you've mixed flasher state and flash process result, suggest two enums: FlasherState, and FlashProcessResult.

patrickelectric: Removed unused errors enums and created a single enum and an error propriety.

src/network/networkmanager.h
```diff
@@ -32,6 +32,14 @@ class NetworkManager : public QObject
      */
     static void requestJson(const QUrl& url, std::function<void(QJsonDocument&)> function);
 
+    /**
+     * @brief Download a file and save in a temporary one
+     *
+     * @param url File to be downloaded
+     * @param function Temporary file path
```
jaxxzer: @param function Callback supplied with file path of downloaded file

src/sensor/sensor.h
```diff
@@ -113,6 +121,9 @@ class Sensor : public QObject
     bool _connected;
     ProtocolDetector* _detector;
     QThread _detectorThread;
+    // For now this will be structures by: name, baudrate, path
```
jaxxzer: you've got three identifiers here, but how does this fit into the key, value?

src/sensor/sensor.h
```diff
@@ -94,11 +95,27 @@ class Sensor : public QObject
      */
     QThread* detectorThread() { return &_detectorThread; };
 
+
+    /**
+     * @brief Return flasher class used by this sensor
+     *  TODO: This should be moved to a singleton flasher instance
```
jaxxzer: What is the reason it should be a singleon? What if there is more than one device connected or more than one type of device connected?

qml/FirmwareUpdate.qml
```diff
@@ -161,7 +161,7 @@ Item {
     PingFileDialog {
```
jaxxzer: commit typo `shotcuts`

qml/FirmwareUpdate.qml
```diff
@@ -40,88 +55,99 @@ Item {
...
+                        Label {
+                            id: firmwareLabel
+                            text: "Firmware File: " + fileDialog.fileName
+                            Layout.fillWidth: true
+                            horizontalAlignment: Text.AlignRight
+                            visible: automaticUpdateCB.currentIndex === 1
+                        }
+
+                        Label {
+                            text: "Last version: "
```
jaxxzer:
```suggestion
                            text: "Latest version: "
```

src/flash/flasher.cpp
```diff
@@ -0,0 +1,151 @@
...
+QString Flasher::stm32flashPath()
+{
+#ifdef Q_OS_OSX
+    // macdeployqt file do not put stm32flash binary in the same folder of pingviewer
+    static QString binPath = QCoreApplication::applicationDirPath() + "/../..";
+#else
+    static QString binPath = QCoreApplication::applicationDirPath();
+#endif
+#ifdef Q_OS_WIN
+    return binPath + "/stm32flash.exe";
```
jaxxzer: `_binRelativePath`?

src/flash/flasher.h
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * @brief Manage the project Flasher
+ *
+ */
+class Flasher : public QObject
+{
+    Q_OBJECT
+public:
+    /**
+     * @brief Construct a new Network Manager object
```
jaxxzer:
```suggestion
     * @brief Construct a new Flasher object
```

src/flash/flasher.h
```diff
@@ -0,0 +1,111 @@
...
+     *
+     */
+    Flasher(QObject* parent = nullptr);
+
+    /**
+     * @brief Destroy the Flash Manager object
```
jaxxzer:
```suggestion
     * @brief Destroy the Flasher object
```

src/flash/flasher.h
```diff
@@ -0,0 +1,111 @@
...
+     *
+     */
+    ~Flasher() = default;
+
+    /**
+     * @brief Set the error type and message
+     *  this function will emit an error signal and update the error propriety
```
jaxxzer:
```suggestion
     *  this function will emit an error signal and update the error property
```

src/flash/flasher.h
```diff
@@ -0,0 +1,111 @@
...
+     */
+    void setError(const QString& errorMessage);
+
+    /**
+     * @brief Return error string
+     *
+     * @return QString
+     */
+    QString error() const { return _error; };
+    // Error will not be updated by the QML par
```
jaxxzer:
```suggestion
    // Error will not be updated by the QML part
```

src/sensor/sensor.h
patrickelectric: Then comment was wrong, it was written during the development.

src/sensor/sensor.h
patrickelectric: Because flasher is a service as the others singletons that we have,
I dont see the relation of multiple devices and the flasher class as a singleton.

jaxxzer: What if you want to flash two devices at once? What if you want to flash an avr or microchip instead of stm?

patrickelectric: > What if you want to flash two devices at once?

This will not work with multiple flasher instances, since stm32flash binary or the serial port can be locked by others instances. 

> What if you want to flash an avr or microchip instead of stm?

That's why we have the flasher class, to abstract the flash method for each hardware/solution.

src/flash/flasher.cpp
jaxxzer: Oy, we're not using this https://github.com/bluerobotics/ping-viewer/pull/361/files#diff-21335bea06bc2e8e55ecdd147181ff74R17

It's redundant here, no?

PR #359, opened 2018-12-13
[WIP] Update interface and add ping enable and "set device id"

patrickelectric:
Set device id is not working !
Fix https://github.com/bluerobotics/ping-viewer/issues/319
Fix https://github.com/bluerobotics/ping-viewer/issues/320

PR #363, opened 2018-12-18
Update main configuration interface

patrickelectric:
Follow this changes: https://github.com/bluerobotics/ping-viewer/pull/359#issuecomment-447079705

### DIFF
--- a/src/settings/settingsmanager.h
+++ b/src/settings/settingsmanager.h
@@ -109,6 +109,8 @@ private:
     AUTO_PROPERTY(bool, replayMenu, false)
     AUTO_PROPERTY(bool, reset, false)
     AUTO_PROPERTY(bool, darkTheme, false)
+    AUTO_PROPERTY(bool, enableConnectionOptions, false)
+    AUTO_PROPERTY(bool, enableSensorAdvancedConfiguration, false)
     //AUTO_PROPERTY_MODEL(QString, adistanceUnits, QStringList, MODEL({"Metric", "Imperial"})) // Example
     AUTO_PROPERTY_JSONMODEL(distanceUnits, QByteArrayLiteral(R"({
             "settings": [
--- a/qml/InfoPage.qml
+++ b/qml/InfoPage.qml
@@ -310,7 +310,7 @@ Item {
         id: popup
 
         GroupBox {
-            id: connGroup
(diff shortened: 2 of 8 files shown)


src/settings/settingsmanager.h
```diff
@@ -109,6 +109,8 @@ class SettingsManager : public QObject
     AUTO_PROPERTY(bool, replayMenu, false)
     AUTO_PROPERTY(bool, reset, false)
     AUTO_PROPERTY(bool, darkTheme, false)
+    AUTO_PROPERTY(bool, connectionGroup, false)
```
jaxxzer: showConnectionOptions?

src/settings/settingsmanager.h
```diff
@@ -109,6 +109,8 @@ class SettingsManager : public QObject
...
+    AUTO_PROPERTY(bool, sensorAdvancedConfiguration, false)
```
jaxxzer: showAdvancedSensorOptions?

src/sensor/ping.h
```diff
@@ -59,7 +59,46 @@ class Ping : public Sensor
...
      */
     uint8_t srcId() { return _srcId; }
-    Q_PROPERTY(int srcId READ srcId NOTIFY srcIdUpdate)
+
+    /** TODO: This is not working
+     * @brief Set device id
+     *
+     * @param id
+     */
+    void srcId(int id)
```
jaxxzer: I think maybe we should refactor this to device_id

srcID is in context of a message. Here we are representing state of a device.

jaxxzer:
I think it will be better if the effect of the enable/disable click can be immediately realized. Right now, the user enables it, then has to go find it. If the option and the effect are in the same visual context, it will be more clear and easy to use.

Perhaps a dropdown/expand menu sort of arrow with heading:

Ping main settings
[ always show]
Ping advanced settings >
Connection settings > <--- click row/arrow to expand menu immediately below.

src/sensor/ping.h
patrickelectric: oooops, this should not be here.

patrickelectric:
@jaxxzer 	new layout.
deepin-screen-recorder_select area_20181219092659

PR #366, opened 2018-12-19
Use PingFileDialog and do some random corrections

patrickelectric:


qml/PingFileDialog.qml
```diff
@@ -0,0 +1,45 @@
...
+    signal accepted(var file)
+
+    FileDialog {
+        id: fileDialog
+        //title: "Please choose a log file"
+        //nameFilters: ["Binary files (*.bin)"]
+        folder: shortcuts.home
+        visible: root.visible
+        onAccepted: {
+            var sizeToRemove = fileDialog.folder.toString().length - fileDialog.fileUrl.toString().length
```
jaxxzer: Can you explain why it is necessary to do all this string manipulation on the file url instead of just using the file url? This seems like a lot of work and extra code for a straightforward use case. I know QUrls are not exactly straightforward to work with, but I think that is what you are supposed to do when handling files with Qt.

patrickelectric: This is based in the [QUrl implementation](https://code.woboq.org/qt5/qtbase/src/corelib/io/qurl.cpp.html#_ZNK11QUrlPrivate11toLocalFileE12QUrlTwoFlagsIN4QUrl19UrlFormattingOptionENS1_25ComponentFormattingOptionEE). 

The qml Url time is not QUrl, so functions like `path()`, `toLocalPath()` and etc does not exist.

You can check more about it [here](https://stackoverflow.com/questions/38610346/local-file-representation-of-qmls-url-type/38619149),  and [here](https://bugreports.qt.io/browse/QTBUG-54988).

patrickelectric: Also, part of this is to only get the filename and other to get the `toLocalPath`

patrickelectric: I updated the code with this comments and organizing it a bit more.

PR #368, opened 2018-12-20
Enable ping

patrickelectric:


qml/MainMenu.qml
```diff
@@ -206,6 +206,13 @@ Item {
                             }
                         }
                     }
+
+                    Switch {
+                        text: "Ping Enabled"
+                        Layout.columnSpan: 1
+                        checked: ping.pingEnable
+                        onCheckedChanged: ping.pingEnable = checked
```
jaxxzer: Should we nop if `ping.pingEnable == checked`? I'm often getting myself into strange signal loops, or multiple events with these qml handlers.

patrickelectric: We don't have recursive signals in this loops, we only set the values when reading from ping.
But to do that we need to make sure that the variables are in populated state and this state represents the sensor variable. 
So, all variables should have at least 2 components, one that it says that's populated and one that shows the value.

patrickelectric: Can you give an example of this "signals loops" in the interface ?
Probably they are from elsewhere but not in the sensor AFAIK.

jaxxzer: I don't know of any in our application, I mean when I am writing qml myself, I usually put guards like this to avoid it because I encounter it when I don't expect sometimes.

jaxxzer: Here is what I mean to try to point out:
- user clicks checkbox
- oncheckedchanged: set ping.enable
- ping.enable changes - checked property of qml is bound to this
- oncheckedchanged: does this run again? I suppose the checked state should not have changed, but it has been updated in the last step.

patrickelectric: And we have this tests, like:
https://github.com/bluerobotics/ping-viewer/blob/master/src/settings/settingsmanagerhelper.h#L52

https://github.com/bluerobotics/ping-viewer/blob/master/src/style/stylemanager.h#L18

And others places where this is necessary.
The problem with the ping class is that we can't make sure that the modification is done in the sensor.

and this is not correct:

> * user clicks checkbox
> * oncheckedchanged: set ping.enable
> * ping.enable changes - checked property of qml is bound to this
> * oncheckedchanged: does this run again? I suppose the checked state should not have changed, but it has been updated in the last step.

What is happening is:

* user clicks checkbox
* oncheckedchanged: set ping.enable
* waits for sensor answer from set_ping_enable request
* if answer is the same one in the widget
	* ping.enable changes (after the sensor answer)- checked property of qml is bound to this
	* oncheckedchanged: do not run again
* if answer is not the same one in the widget
	* ping.enable changes (after the sensor answer)- checked property of qml is bound to this
    * oncheckedchanged: will run again and a ping_set_enable will be called with the actual state, after that nothing will change

The problem is, the other way around may result in poor user experience, since the controls will fail to show the actual state of the sensor. This was a problem in the pass.

The way to fix that, is to do how it was said here:
https://github.com/bluerobotics/ping-viewer/pull/368#discussion_r243403333

Or anything similar that ensures that the state of the sensor in synchronized with the sensor and ensures that avoiding any crazy loop.

PR #382, opened 2019-01-15
Ping1DVisualizer: Hide readout line when mouse not in chart area

Williangalvani:
hidebar2

### DIFF
--- a/qml/Ping1DVisualizer.qml
+++ b/qml/Ping1DVisualizer.qml
@@ -51,6 +51,7 @@ Item {
                 height: 15
                 width: waterfall.height
                 transform: Rotation { origin.x: height/2; angle: 90}
+                visible: waterfall.containsMouse
                 gradient: Gradient {
                     GradientStop { position: 0.3; color: "transparent" }
                     GradientStop { position: 0.5; color: StyleManager.secondaryColor } // Not working with material
--- a/src/waterfall/waterfall.cpp
+++ b/src/waterfall/waterfall.cpp
@@ -20,6 +20,7 @@ Waterfall::Waterfall(QQuickItem *parent):
     _maxDepthToDrawInPixels(0),
     _minDepthToDrawInPixels(0),
     _mouseDepth(0),
+    _containsMouse(false),
     _smooth(true),
     _updateTimer(new QTimer(this)),
     currentDrawIndex(displayWidth)
@@ -397,5 +398,15 @@ void Waterfall::hoverLeaveEvent(QHoverEvent *event)
 {
     Q_UNUSED(event)
     emit mouseLeave();
+    // The mouse is not inside the waterfall area, so set the depth under the mouse to an invalid value
     _mouseDepth = -1;
+    _containsMouse = false;
+    emit containsMouseChanged();
+}
+
+void Waterfall::hoverEnterEvent(QHoverEvent *event)
+{
+    Q_UNUSED(event)
+    _containsMouse = true;
+    emit containsMouseChanged();
 }
--- a/src/waterfall/waterfall.cpp
+++ b/src/waterfall/waterfall.cpp
@@ -397,7 +397,6 @@ void Waterfall::hoverMoveEvent(QHoverEvent *event)
 void Waterfall::hoverLeaveEvent(QHoverEvent *event)
 {
     Q_UNUSED(event)
-    emit mouseLeave();
     // The mouse is not inside the waterfall area, so set the depth under the mouse to an invalid value
     _mouseDepth = -1;
     _containsMouse = false;
--- a/src/waterfall/waterfall.h
+++ b/src/waterfall/waterfall.h
@@ -62,6 +62,14 @@ public:
     Q_PROPERTY(QPoint mousePos READ mousePos NOTIFY mousePosChanged)
 
     /**
+     * @brief Return true if the mouse is currently inside the waterfall area
+     *
+     * @return bool
+     */
+    bool containsMouse() {return _containsMouse;}
+    Q_PROPERTY(bool containsMouse READ containsMouse NOTIFY containsMouseChanged)
+
+    /**
      * @brief Get theme name used in the waterfall
      *  Check WaterfallGradient
      *
@@ -128,6 +136,7 @@ public:
     float _mouseColumnConfidence;
     float _mouseColumnDepth;
     float _mouseDepth;
+    bool _containsMouse;
     QPoint _mousePos;
     bool _smooth;
     QTimer* _updateTimer;
@@ -158,9 +167,11 @@ signals:
     void mouseDepthChanged();
     void mouseColumnConfidenceChanged();
     void mouseColumnDepthChanged();
+    // TODO: mouseMove should be renamed
     void mouseMove();
     void mousePosChanged();
     void mouseLeave();
+    void containsMouseChanged();
     void themeChanged();
     void themesChanged();
     void smoothChanged();
@@ -244,11 +255,18 @@ public:
     void hoverMoveEvent(QHoverEvent *event);
 
     /**
-     * @brief Function that deals when the mouse is outside the waterfall
+     * @brief Function that executes when the mouse leaves the waterfall area
      *
      * @param event
      */
     void hoverLeaveEvent(QHoverEvent *event);
+
+    /**
+     * @brief Function that executes when the mouse enters the waterfall area
+     *
+     * @param event
+     */
+    void hoverEnterEvent(QHoverEvent *event);
 
 private:
     Q_DISABLE_COPY(Waterfall)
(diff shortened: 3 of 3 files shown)


qml/Ping1DVisualizer.qml
```diff
@@ -91,6 +92,14 @@ Item {
                 end_mm: waterfall.maxDepthToDraw
                 visible: start_mm != end_mm
             }
+
+            onMouseLeave: {
+                readoutLine.visible = false
+            }
+
+            onMouseMove: {
```
patrickelectric: Since waterfall is a c++ class, you could create a mouseInside variable and do:
`readoutLine.visible = waterfall.mouseInside`

Williangalvani: done

src/waterfall/waterfall.cpp
```diff
@@ -37,6 +38,13 @@ Waterfall::Waterfall(QQuickItem *parent):
     connect(_updateTimer, &QTimer::timeout, this, [&] {update();});
     _updateTimer->setSingleShot(true);
     _updateTimer->start(50);
+    connect(this, &Waterfall::mouseLeave, this, [this] {_mouseInside = false; emit mouseInsideChanged();});
+    connect(this, &Waterfall::mouseMove, this, [this] {
```
jaxxzer: Will this introduce unnecessary overhead?

jaxxzer: Where is are these mouseLeave and mouseMove signals documented? Is there a mouseEnter? Is this route really better than the qml @Williangalvani had up originally?

patrickelectric: He could use the function that already exist..
mouseLeave and mouseMove are signals from this class, he could set the `mouseInside` variable inside this two functions and do the emit.

This will make the waterfall class more complete as widget and will make the qml part much more simple.

jaxxzer: Whoops, I thought they were inherited from QQuickPaintedItem.

Williangalvani: I think I got it right this time.

src/waterfall/waterfall.cpp
```diff
@@ -398,4 +399,13 @@ void Waterfall::hoverLeaveEvent(QHoverEvent *event)
     Q_UNUSED(event)
     emit mouseLeave();
     _mouseDepth = -1;
+    _mouseInside = false;
+    emit mouseInsideChanged();
+}
+
+void Waterfall::hoverEnterEvent(QHoverEvent *event)
+{
+    Q_UNUSED(event)
```
jaxxzer: should we emit a `mouseEnter()` here?
do we need to do anything with `_mouseDepth`?

Williangalvani:
I renamed the property to `containsMouse` and removed the unused `mouseLeave()` signal

src/waterfall/waterfall.cpp
```diff
@@ -396,6 +397,14 @@ void Waterfall::hoverMoveEvent(QHoverEvent *event)
 void Waterfall::hoverLeaveEvent(QHoverEvent *event)
 {
     Q_UNUSED(event)
-    emit mouseLeave();
     _mouseDepth = -1;
```
jaxxzer:
```suggestion
    // TODO why, is this necessary?
    _mouseDepth = -1;
```

src/waterfall/waterfall.h
```diff
@@ -160,7 +169,7 @@ class Waterfall : public QQuickPaintedItem
     void mouseColumnDepthChanged();
     void mouseMove();
```
jaxxzer: // TODO mouseMove and mousePosChanged seems redundant, clean up

src/waterfall/waterfall.cpp
patrickelectric: MouseDepth can be used to get the depth where the mouse is pointing.

patrickelectric: The doxygen documentation:
```
     /**
     * @brief Get depth from mouse position
     *
     * @return float
      */
    float mouseDepth() {return _mouseDepth;}
    Q_PROPERTY(float mouseDepth READ mouseDepth NOTIFY mouseDepthChanged)
```

Williangalvani: I couldn't find usages of `mouseDepth`, but i found of `mouseColumnDepth`.
Are they different things?
And are both necessary?

patrickelectric: The doxygen documentation says:
```cpp
/**
 * @brief Get depth from mouse position
 *
 * @return float
 */
float mouseDepth() {return _mouseDepth;}
Q_PROPERTY(float mouseDepth READ mouseDepth NOTIFY mouseDepthChanged)

/**
 * @brief Get signal deepth from mouse position column
 *
 * @return float
 */
float mouseColumnDepth() {return _mouseColumnDepth;}
Q_PROPERTY(float mouseColumnDepth READ mouseColumnDepth NOTIFY mouseColumnDepthChanged)
```

The first one is the mouse position depth, the second one is the mouse column depth

jaxxzer: @patrickelectric I mean why is it necessary to set to -1?

patrickelectric: If mouse is not in the application, it sets the depth to an invalid value.

jaxxzer: So if mouse enters application, do we need to reinitialize to a valid value?

src/waterfall/waterfall.h
patrickelectric: They are not the same thing.

src/waterfall/waterfall.cpp
patrickelectric: This is done automatically by `hoverMoveEvent`

src/waterfall/waterfall.h
jaxxzer: English indicates otherwise: 'move' is the same as 'delta position'
We need comments or better identifiers.

patrickelectric: Yes, mouseMove should be renamed to mouseDepth

Williangalvani:
I rebased and added a TODO to rename `mouseMove()`

src/waterfall/waterfall.h
```diff
@@ -128,6 +136,7 @@ class Waterfall : public QQuickPaintedItem
     float _mouseColumnConfidence;
     float _mouseColumnDepth;
     float _mouseDepth;
+    bool _containsMouse;
```
patrickelectric: Alphabetic order

src/waterfall/waterfall.cpp
jaxxzer:
```suggestion
    // The mouse is not inside the application, so set the depth under the mouse to an invalid value
    _mouseDepth = -1;
```

src/waterfall/waterfall.h
```diff
@@ -250,6 +260,13 @@ class Waterfall : public QQuickPaintedItem
      */
     void hoverLeaveEvent(QHoverEvent *event);
 
+    /**
+     * @brief Function that deals when the mouse is inside the waterfall
```
jaxxzer:
```suggestion
     * @brief Function that deals when the mouse enters the waterfall area
```

src/waterfall/waterfall.h
```diff
@@ -61,6 +61,14 @@ class Waterfall : public QQuickPaintedItem
     QPoint mousePos() {return _mousePos;}
     Q_PROPERTY(QPoint mousePos READ mousePos NOTIFY mousePosChanged)
 
+    /**
+     * @brief Return true if the cursor is currently inside the waterfall area
```
jaxxzer:
```suggestion
     * @brief Return true if the mouse is currently inside the waterfall area
```

(consitency)

src/waterfall/waterfall.h
Williangalvani: I'd rather keep mouse-related variables together. And the rest is also pretty much out of order. I don't mind making a commit to organize that, but seems out of scope for this PR.

jaxxzer: https://github.com/bluerobotics/ping-viewer/issues/411

We've got a lot TODO already, I think we can move along now that we have notes of everything.

src/waterfall/waterfall.h
```diff
@@ -244,12 +254,19 @@ class Waterfall : public QQuickPaintedItem
     void hoverMoveEvent(QHoverEvent *event);
 
     /**
-     * @brief Function that deals when the mouse is outside the waterfall
+     * @brief Function that executes when the mouse enters the waterfall area
```
jaxxzer: @Williangalvani You made a mistake. :)

jaxxzer:
@Williangalvani so close, three to resolve.

src/waterfall/waterfall.h
Williangalvani: :man_facepalming:

jaxxzer:
Waiting @patrickelectric approval.

patrickelectric:
@Williangalvani 	Please rebase to be merged,

PR #384, opened 2019-01-15
Correct open log path

patrickelectric:


src/filemanager/filemanager.h
```diff
@@ -56,6 +57,14 @@ class FileManager : public QObject
      */
     QFileInfoList getFilesFrom(Folder folderType);
 
+    /**
+     * @brief Get the path used by the program to this file type
+     *
+     * @param folderType
+     * @return QUrl
+     */
+    Q_INVOKABLE QUrl getPathFrom(FileManager::Folder folderType);
```
jaxxzer: getPath**To**?

patrickelectric: I just the name of the last function to get things _From_ Folder, like path, files and etc. get*From and get*To in the same API will make it a bit confuse.

PR #386, opened 2019-01-16
Check firmware HEX file before flash

patrickelectric:


src/sensor/hexvalidator.cpp
```diff
@@ -0,0 +1,93 @@
...
+    // Line Format:
+    //      :AABBBBCCnXDD
+    // AA: Byte count
+    // BBBB: Address
+    // CC: Record type
+    // nX: n characters, where n is AA*2
+    //  - If AA is 3, nX will be: 0102FF
+    // DD: Checksum byte
+
+    auto error = [&bytes](const char* text, ...) {
```
jaxxzer: nice

PR #418, opened 2019-01-21
main: Remove hash with normal releases

### LINKED ISSUE #417: Application version - display tag only if commit is tagged
The `t0.8.5` tag is `8b72502` - The version in the application should just be `t0.8.5` as `t0.8.5-8b72502` is redundant.

The **newer** continuous release is `39c4031` - The version in the application bar should just say `39c4031` or `continuous-39c4031` - The commit is not tagged, so we should not show a tag. We are currently showing `t0.8.5` on the version here, and I think that is misleading and could lead to some confusion in support.

patrickelectric:
Fix #417

### DIFF
--- a/pingviewer.pro
+++ b/pingviewer.pro
@@ -66,7 +66,7 @@ message("Check for GIT version and tag:")
 exists ($$_PRO_FILE_PWD_/.git) {
     GIT_VERSION = $$system(git --git-dir $$_PRO_FILE_PWD_/.git --work-tree $$PWD log -1 --format=%h)
     GIT_VERSION_DATE = $$system(git --git-dir $$_PRO_FILE_PWD_/.git --work-tree $$PWD log -1 --format=%aI)
-    GIT_TAG = $$system(git --git-dir $$_PRO_FILE_PWD_/.git --work-tree $$PWD describe --abbrev=0 --tags --exclude continuous)
+    GIT_TAG = $$system(git --git-dir $$_PRO_FILE_PWD_/.git --work-tree $$PWD describe --tags --exclude "continuous*" --exact-match)
     GIT_URL = $$system(git --git-dir $$_PRO_FILE_PWD_/.git --work-tree $$PWD remote get-url origin)
     DEFINES += 'GIT_VERSION=\\"$$GIT_VERSION\\"'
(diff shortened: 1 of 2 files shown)


qml/main.qml
```diff
@@ -39,6 +39,11 @@ ApplicationWindow {
         var tag = GitTag === "" ? "No tags!" : GitTag
         var hash = GitVersion
 
-        return  applicationName + " - " + tag + " - " + hash
+        var title = applicationName + " - " + tag
+        if(tag.includes("continuous")) {
```
jaxxzer: This tag 'continuous' will not exist on local development. Also, the tag will never be "" with the current `git describe` command. (and if it did, the application version would say pingviewer - No Tags! without any hash).

I've pushed some commits to handle this.

pingviewer.pro
```diff
@@ -66,7 +66,7 @@ message("Check for GIT version and tag:")
 exists ($$_PRO_FILE_PWD_/.git) {
     GIT_VERSION = $$system(git --git-dir $$_PRO_FILE_PWD_/.git --work-tree $$PWD log -1 --format=%h)
     GIT_VERSION_DATE = $$system(git --git-dir $$_PRO_FILE_PWD_/.git --work-tree $$PWD log -1 --format=%aI)
-    GIT_TAG = $$system(git --git-dir $$_PRO_FILE_PWD_/.git --work-tree $$PWD describe --abbrev=0 --tags --exclude continuous)
+    GIT_TAG = $$system(git --git-dir $$_PRO_FILE_PWD_/.git --work-tree $$PWD describe --tags --exclude continuous* --exact-match)
```
patrickelectric: @jaxxzer 	how can this work with normal releases ?

patrickelectric: What is idea behind this change ? Why do you want to remove normal and test releases ?

qml/main.qml
patrickelectric: We are not going to deploy with "local development".
git describe can fail if you are running a old git version.
The best way to fix the case where you don't have a tag is to use regex to find the release type.

pingviewer.pro
patrickelectric: Fixed

PR #419, opened 2019-01-21
Improve todo notation

patrickelectric:
Use `TODO:` and not `todo`, `todo,` or any variation


jaxxzer:
We should add this to the guideline doc.

jaxxzer:
My preference would be `TODO` not `TODO:`.

patrickelectric:
In most IDEs the keyword are TODO:, FIXME:, DEBUG:, and etc. I never saw one without the two dots in the end.

PR #428, opened 2019-01-25
Update pingviewer icon

patrickelectric:
Based on @rjehangir work.
image


rjehangir:
I like it but I'm open to feedback and other opinions. Let me know!

jaxxzer:
Does this change the icon in the upper left of the application info menu? (I think it should not)

patrickelectric:
@jaxxzer 	Are you talking about the info menu ? Nops.

PR #430, opened 2019-01-25
Add outline in texts that are over the waterfall

patrickelectric:
Before:
image

After:
image


rjehangir:
Nice! I'd recommend an outline on the red bottom-tracking arrow as well.

patrickelectric:
@rjehangir 	done !

PR #436, opened 2019-01-28
Rework normal/advanced ping settings

### LINKED ISSUE #414: Reorganize 'simple' and 'advanced' Ping device settings
The current organization is:

## Menu Items

- **Emit Ping:** Request a single profile sample from the device, and stop continuous transmission of profile data (if active).
- **Ping Rate Slider:** Enable continuous transmission of profile data, and set the maximum update rate (the actual update rate may be slower).
- **Auto Gain:** Enable automatic gain and range adjustments (ENABLED is the recommended default settings).
- **Gain Selection:** Manual gain control (automatic mode must be disabled). Adjust the sensitivity of the device; larger numbers indicate greater sensitivity.

#### Advanced Settings

Click the wrench on the Device Settings menu to access these advanced settings
- **Speed of Sound:** The speed of sound to be used for distance calculations. This should be 1500 m/s in salt water, 1435 m/s in fresh water, and 343 m/s in air (for testing!).
- **Ping Enabled:** Enable the acoustic output of the Ping device. When set to disabled, the Ping will not emit energy from the transducer. When set to enabled, energy may be emitted for background operations, even when there is no data being transmitted.
- **Scan Range:** Manual scan range selection (automatic mode must be disabled). Manually select the scan range for data output and target tracking. There are two parameters to adjust the scan range:
 - *Scan Start*: The distance away from the Ping device transducer to begin the scan
 - *Scan Length*: The total distance to be scanned (the distance from *Scan Start* to end the scan)

***

I think a better organization is:

## Menu Items

- **Emit Ping:** Request a single profile sample from the device, and stop continuous transmission of profile data (if active).
- **Ping Rate Slider:** Enable continuous transmission of profile data, and set the maximum update rate (the actual update rate may be slower).
- **Speed of Sound:** The speed of sound to be used for distance calculations. This should be 1500 m/s in salt water, 1435 m/s in fresh water, and 343 m/s in air (for testing!).
- **Ping Enabled:** Enable the acoustic output of the Ping device. When set to disabled, the Ping will not emit energy from the transducer. When set to enabled, energy may be emitted for background operations, even when there is no data being transmitted.

#### Advanced Settings

Click the wrench on the Device Settings menu to access these advanced settings

- **Auto Gain:** Enable automatic gain and range adjustments (ENABLED is the recommended default settings).
- **Gain Selection:** Manual gain control (automatic mode must be disabled). Adjust the sensitivity of the device; larger numbers indicate greater sensitivity.
- **Scan Range:** Manual scan range selection (automatic mode must be disabled). Manually select the scan range for data output and target tracking. There are two parameters to adjust the scan range:
 - *Scan Start*: The distance away from the Ping device transducer to begin the scan
 - *Scan Length*: The total distance to be scanned (the distance from *Scan Start* to end the scan)

jaxxzer:
Fix #414 
image

### DIFF
--- a/qml/MainMenu.qml
+++ b/qml/MainMenu.qml
@@ -65,7 +65,7 @@ Item {
                     spacing: 2
 
                     Text {
-                        text: "Ping Enabled"
+                        text: "Ping Enabled:"
                         color: Material.primary
                     }
 
@@ -140,7 +140,7 @@ Item {
                     visible: SettingsManager.enableSensorAdvancedConfiguration
 
                     Text {
-                        text: "Auto Gain"
+                        text: "Auto Gain:"
                         color: Material.primary
                     }
 
--- a/qml/MainMenu.qml
+++ b/qml/MainMenu.qml
@@ -64,8 +64,12 @@ Item {
                 RowLayout {
                     spacing: 2
 
-                    Switch {
+                    Text {
                         text: "Ping Enabled"
+                        color: Material.primary
+                    }
+
+                    Switch {
                         Layout.columnSpan: 1
                         checked: ping.pingEnable
                         onCheckedChanged: ping.pingEnable = checked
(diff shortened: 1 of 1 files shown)


patrickelectric:
The gain combobox is huge

patrickelectric:
The combobox should be inside the Auto Gain Text element to avoid the space that the layout is adding it. Also I would believe that `Auto Gain:` with two dots will make more since, since the component now is two different ones. The same thing applies to `Ping Enabled:`.

jaxxzer:
I've tried to do your suggestion about spacing
The checkbox item margins prevent what you are requesting. It will require a custom component. I think it looks ok for v1.

jaxxzer:
I did add the colon `:`

patrickelectric:
Closed by: https://github.com/bluerobotics/ping-viewer/pull/437

Issue #443, opened 2019-01-29
Stable version is not a release!


jaxxzer:
This is on v1.0.0. :man_facepalming:

![image](https://user-images.githubusercontent.com/8315108/51881801-72145d00-234a-11e9-899f-2bbf70450b94.png)

Related to #442 probably.

patrickelectric:
![image](https://user-images.githubusercontent.com/1215497/51909725-34a0e580-23c5-11e9-9490-a92f01d57900.png)
Tested and working with last stable

PR #446, opened 2019-01-29
project: use git tag instead of git describe in order to find all

### LINKED ISSUE #442: Show semver version also for stable release
This is probably happening because we now have two tags with pointing to the same OID.
image

I think we could do `stable-vX.X.X` or just `vX.X.X`.

jaxxzer:
fix #442 
This will show only commits with tags that do not contain 'continuous' or 'stable' in the name.


jaxxzer:
fix #443 
first commit fixes `this is not a release!`
second commit fixes `new version available`

PR #448, opened 2019-01-29
Readout position

### LINKED ISSUE #441: Initialize depth readout position in lower left corner
Here is what it looks like after resetting settings:
image

jaxxzer:
fix #441image


jaxxzer:
This only affects first start, settings still works ok.

PR #456, opened 2019-02-04
Fix protocol detector scan call

### LINKED ISSUE #457: Calling ProtocolDetector::scan directly freezes the interface

patrickelectric:
Fix #457


src/sensor/protocoldetector.cpp
```diff
@@ -26,9 +26,11 @@ ProtocolDetector::ProtocolDetector()
         {LinkType::Udp, {"192.168.2.2", "9090"}, "BlueRov2 standard connection"},
         {LinkType::Udp, {"127.0.0.1", "1234"}, "Development port"}
     });
+
```
jaxxzer:
```suggestion

//doing this because it will freeze the main thread if called directly
```

PR #472, opened 2019-02-20
Fix wrong input when input loses focus

patrickelectric:
TextField with validators does not emit onAccept or onEditingFinished when wrong inputs are in the TextField. We need to check if the focus was lost and the input is valid manually.

onEditingFinished is only emitted when TextField has focus, otherwise will not emit. That's why I'm moving to on Accept, this one will be emitted if the application loses focus or enter is emitted. 

Input variable from onEditingFinished was removed to use old/classic implementation of text variable after fixing logic problems with new implementation.

### DIFF
--- a/qml/ConnectionMenu.qml
+++ b/qml/ConnectionMenu.qml
@@ -184,9 +184,10 @@ Item {
                     Layout.columnSpan:  2
                     Layout.fillWidth: true
                     onEditingFinished: {
-                        if (input == "0.0.0.0" || input == "localhost") {
+                        if (text === "0.0.0.0" || text === "localhost") {
                             text = "127.0.0.1"
                         }
+
                         connect(AbstractLinkNamespace.Udp, text, udpPort.text)
                     }
                 }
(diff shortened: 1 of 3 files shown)


patrickelectric:
@jaxxzer 	^

qml/ConnectionMenu.qml
```diff
@@ -184,10 +184,11 @@ Item {
                     Layout.columnSpan:  2
                     Layout.fillWidth: true
                     onEditingFinished: {
-                        if (input == "0.0.0.0" || input == "localhost") {
+                        if (text === "0.0.0.0" || text === "localhost") {
```
jaxxzer: should this be `udpIp.text` (as you have on line 191)?

jaxxzer: Or maybe 191 should be just `text`. Looking for consistency or understanding if there is some reason for the difference.

qml/PingTextField.qml
```diff
@@ -5,46 +5,56 @@ import QtQuick.Layouts 1.3
...
-            root.editingFinished(textField.text)
+            // It's necessary to update the input variable
+            root.text = textField.text
+            root.editingFinished()
         }
 
         onActiveFocusChanged: {
+            // TODO: We should update the user about wrong inputs here
+            // This allow us to correct not valid inputs to the last valid value
+            if(!acceptableInput) {
```
jaxxzer: `acceptableInput` is not defined, do you mean to comment this code block along with the comment?

PR #474, opened 2019-02-20
KCrash support for windows

patrickelectric:
**THIS PR NEEDS MODIFICATIONS AFTER ACCEPTING**
[auto crash feature to test PR and deployment]

This adds kcrash on windows deployment.
To test it, this binary will autocrash after some seconds and a window will appear:
image

image

image

### DIFF
--- a/src/logger/loglistmodel.h
+++ b/src/logger/loglistmodel.h
@@ -77,6 +77,26 @@ public:
     QSortFilterProxyModel* filteredModel() { return &_filter; };
     Q_PROPERTY(QSortFilterProxyModel* filteredModel READ filteredModel CONSTANT)
 
+    /**
+     * @brief Do the necessary connections of the object
+     *
+     * This also helps to avoid msvc crashs.
+     * Such crashs happens by unknown reasons,
+     * since we are talking about Mycrosoft compiler,
+     * this unknown reasons are really complex, hard to find and maybe
+     * a creation of a superior being to test ourselfs
+     * in a perpetual and impossible fight against the dark forces.
+     * If you are reading this from the future or using mingw to deploy,
+     * you can try move everything here to the constructor.
+     * Note: If the computer starts to scream something about:
+     * "read access violation at: 0x0, flags=0x0"
+     * I highly advise you to run `git stash && git checkout master`,
+     * and forget everything related to this lines.
+     * And if someone asks you to fix that: https://www.youtube.com/watch?v=5GfOlP8FKt0
+     * Get in touch with me otherwise.
+     */
+    void start();
+
 signals:
     /**
      * @brief Append a new log message in model
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -23,7 +23,7 @@ configuration:
   - debug
 
 install:
-  # Set some vars
+  # Set and load necessary variables
   - '%QTDIR%\bin\qtenv2.bat'
   - pip install jinja2
   - qmake -v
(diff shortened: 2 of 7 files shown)


patrickelectric:
All random corrections are in https://github.com/bluerobotics/ping-viewer/pull/485

jaxxzer:
I downloaded the debug build from the release pages (here), and it looks like the debug symbols were not found on my computer: https://pastebin.com/9izG8mHh

There is no mention of `networktool.cpp`.

Interestingly, the program keeps running and waterfall pumping out data after the 'crash' :).

patrickelectric:
I just tested it here and it's working: https://pastebin.com/4UWnNkyp, and this same zip was tested by Brian.
Did you follow the steps [described here](https://github.com/bluerobotics/ping-viewer/pull/474#issue-254741851) ?
Can you test @Williangalvani.

jaxxzer:
> Did you follow the steps described here ?

The only thing I think I did different was click 'retry' instead of 'ignore'. If I remember correctly, 'retry' was the only one giving me the stack traces.

patrickelectric:
Try 'ignore' option and if does not work, set QT_LOGGING_RULES to `org.kde.drkonqi.debug=true` and start ping-viewer from the terminal, the debug information will be available for us to understand what is going on;

patrickelectric:
Are you unzipping it under a new folder or with one that already contains an old ping version unzipped ?

jaxxzer:
> Are you unzipping it under a new folder or with one that already contains an old ping version unzipped ?

It went to a new folder with `(1)` suffix. I did also rename it to remove the `(1)` suffix, but with no different result.

Williangalvani:
Works ok here
https://pastebin.com/SNHaUfF8

jaxxzer:
I did your advice @patrickelectric and now I've got .cpp line numbers. I'm not sure it's right though, it doesn't look exactly like yours: https://pastebin.com/aMJxpGPC

For example, my output doesn't mention anything about `pingviewer!QList<QString>::operator[]+0x4b [c:\qt\5.12.1\msvc2017_64\include\qtcore\qlist.h @ 553]` like @Williangalvani 's does.

patrickelectric:
What you did different ?
`QT_LOGGING_RULES` should not change the result. only the debug messages in the terminal. 

About the QString information, that's interesting but I'm not sure why though.

jaxxzer:
> What you did different ?

I did just the two things in your suggestion with `set` and click ignore, but it looks like that did not matter.

Now with the variable unset and click ignore or click retry, I get the right line number @45. I can't reproduce my initial situation.

patrickelectric:
This also updates Qt version to 5.12

src/logger/loglistmodel.h
```diff
@@ -77,6 +77,26 @@ class LogListModel : public QAbstractListModel
...
+     * If you are reading this from the future or using mingw to deploy,
+     * you can try move everything here to the constructor.
+     * Note: If the computer starts to scream something about:
+     * "read access violation at: 0x0, flags=0x0"
+     * I highly advise you to run `git stash && git checkout master`,
+     * and forget everything related to this lines.
+     * And if someone asks you to fix that: https://www.youtube.com/watch?v=5GfOlP8FKt0
+     * Get in touch with me otherwise.
+     */
+    void start();
```
jaxxzer: effing lol

src/network/networktool.cpp
```diff
@@ -41,6 +41,8 @@ void NetworkTool::checkInterfaceUpdate()
 
 void NetworkTool::checkNewVersionInGitHubPayload(const QJsonDocument& jsonDocument)
 {
+    QStringList potato;
+    qDebug() << potato[1e10];
```
jaxxzer: take it out!

src/network/networktool.cpp
```diff
@@ -169,7 +171,7 @@ void NetworkTool::scheduleUpdateCheck()
         timer->deleteLater();
     });
     timer->setSingleShot(true);
-    timer->start(30000);
+    timer->start(10000);
```
jaxxzer: drop the commit

jaxxzer:
awesome!

PR #475, opened 2019-02-21
Add flash feedback for the user

### LINKED ISSUE #469: Add status text in flash widget
https://discuss.bluerobotics.com/t/ping-interrupted-firmware-update/4389/2

patrickelectric:
Fix #469

### DIFF
--- a/src/flash/flasher.h
+++ b/src/flash/flasher.h
@@ -30,21 +30,49 @@ public:
     ~Flasher() = default;
 
     /**
-     * @brief Set the error type and message
-     *  this function will emit an error signal and update the error property
+     * @brief Defines flash state
      *
-     * @param errorMessage
      */
-    void setError(const QString& errorMessage);
+    enum States {
+        Error = -1,
+        Idle,
+        StartingFlash,
+        Flashing,
+        FlashFinished,
+    } states;
+    Q_ENUM(States)
 
     /**
-     * @brief Return error string
+     * @brief Set state message
+     *  this function will emit a message signal and update the message property
+     *
+     * @param message
+     */
+    void setMessage(const QString& message);
+
+    /**
+     * @brief Return message string
      *
      * @return QString
      */
-    QString error() const { return _error; };
-    // Error will not be updated by the QML part
-    Q_PROPERTY(QString error READ error NOTIFY errorChanged)
+    QString message() const { return _message; };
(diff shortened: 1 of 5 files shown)


patrickelectric:
deepin-screen-recorder_select area_20190221171001
deepin-screen-recorder_select area_20190221170731
deepin-screen-recorder_select area_20190221174428

patrickelectric:
@jaxxzer ^

src/flash/flasher.h
```diff
@@ -30,21 +30,34 @@ class Flasher : public QObject
...
+    /**
+     * @brief Set state message
+     *  this function will emit a message signal and update the message property
      *
-     * @param errorMessage
+     * @param message
+     * @param state
      */
-    void setError(const QString& errorMessage);
+    void setMessage(const QString& message, const Flasher::States state);
```
jaxxzer: I think setting the message and setting the state should be independent of each other (we should have a setState).

patrickelectric: The idea is force us to set a text explaining what is happening and to update the user.
Using two different functions to accomplish the same thing will result in missing setMessages or setStates calls.

patrickelectric: Also, I can't imagine a case where setState will be called without a message.

jaxxzer: > The idea is force us to set a text explaining what is happening and to update the user when changing the state.

I understand what is happening here. I don't agree with how you have structured this.

> Using two different functions to accomplish the same thing will result in missing setMessages or setStates calls.

You are doing two different things. You are 1: changing state like a state machine. and 2: providing status/error messages.

Why should a state change depend on a message being emitted? And why should a message being emitted depend on a state change?

I may desire to silently change state without a message. When changing from idle to flashing, there is no need to emit a message saying "I'm beginning to flash this" (this message doesn't convey any additional information than the state change itself; the client can watch the state change happen and come up with the message itself for example!).

I can emit a message while I'm flashing: "1%", "2%"... and the state does not change or need to be updated.

I may also have a handler that cares only about the state and not the message, or about the message and not the state.

The suggestion is to make the interface more flexible, and to decouple error/update messages from the basic function of the state machine.

I've made similar suggestions in the past about this: https://github.com/bluerobotics/ping-viewer/pull/361#discussion_r253523061

patrickelectric:
Add `setState` with message as optional argument, `setMessage` and individual signals for each.

PR #480, opened 2019-03-18
Fix readout position 

### LINKED ISSUE #462: Depth readout disappear when maximizing window

patrickelectric:
Fix #462

### DIFF
--- a/qml/ValueReadout.qml
+++ b/qml/ValueReadout.qml
@@ -6,7 +6,6 @@ import SettingsManager 1.0
 
 Item {
     id: root
-    x: margin
     width: mainLayout.width
     height: mainLayout.height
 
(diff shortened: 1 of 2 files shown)


qml/ValueReadout.qml
```diff
@@ -119,13 +118,24 @@ Item {
...
+    function checkPosition() {
+        // Check if parent is visible and correct initialized
+        if(!root.parent.visible || root.parent.width === 0 || root.parent.height === 0) {
+            return
+        }
+        if(y <= 0 || y + root.height > parent.height) {
+            y = root.parent.height - root.height - margin
         }
-        if(x <= 0 || x + width > parent.width) {
+        if(x <= 0 || x + root.width > parent.width) {
```
Williangalvani: How does this work?
Why is the x = margin on both upper and lower limits?
Wouldn't it make more sense to just limit to upper and lower bounds?
why not just `x = Math.max(Math.min(0, x), parent.height - root.height)`? 
Or even better, make a more readable `clamp(number, min, max)` based on that.

patrickelectric: > Why is the x = margin on both upper and lower limits?

This code is just to avoid an outbound position of readout.

> why not just x = Math.max(Math.min(0, x), parent.height - root.height)?

Yes, this is an alternative.

PR #490, opened 2019-04-03
Replace html images with markdown

Williangalvani:
This adds a css rule to center all images in text, and replaces the html img tags with markdown where possible

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ install:
 script:
   - git remote set-url origin https://${GITHUB_TOKEN}@github.com/bluerobotics/ping-viewer
   - mkdocs build
-  - htmlproofer --empty-alt-ignore --file-ignore "./site/404.html" --url-ignore "https://fonts.gstatic.com" || travis_terminate 1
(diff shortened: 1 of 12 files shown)


patrickelectric:
@Williangalvani remind to squash some of this commits.

.travis.yml
```diff
@@ -9,7 +9,7 @@ install:
 script:
   - git remote set-url origin https://${GITHUB_TOKEN}@github.com/bluerobotics/ping-viewer
   - mkdocs build
-  - htmlproofer --empty-alt-ignore --file-ignore "./site/404.html" --url-ignore "https://fonts.gstatic.com" || travis_terminate 1
+  - htmlproofer --empty-alt-ignore --file-ignore "./site/404.html" --url-ignore "https://fonts.gstatic.com" site || travis_terminate 1
```
jaxxzer: What exactly does this do? Is it a functional change?

Williangalvani: this runs the htmlproofer in what is effectively `/` for the website. Without it, absolute links are broken. Instead of looking for images in `./site/images`(where they are), it tries to find them in `./images` and it fails.

jaxxzer:
:clap:

PR #493, opened 2019-04-05
Revert to relative paths

Williangalvani:
The fact that our docs are hosted in `/ping-viewer/` makes the absolute links not work. I expected mkdocs to be able to handle that, but [they just do not support absolute links](https://github.com/mkdocs/mkdocs/issues/1592).

### DIFF
--- a/docs/application-information.md
+++ b/docs/application-information.md
@@ -6,7 +6,7 @@
 4. [**Application log**](#application-log)
 5. [**Log categories**](#log-categories)
 
-![Application information annotated](/images/viewer/annotated/application-information-annotated.png)
+![Application information annotated](../images/viewer/annotated/application-information-annotated.png)
 
 #### Application software information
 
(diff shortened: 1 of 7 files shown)


jaxxzer:
:man_facepalming:

jaxxzer:
We could also just prefix all the paths with `ping-viewer/` instead of revert.

Williangalvani:
> We could also just prefix all the paths with `ping-viewer/` instead of revert.

I thought about that, but that is going to make html-proofer fail

jaxxzer:
But.. the link would be correct.

And right now, the links are broken, how is htmlproofer passing? Is it checking image links?

Williangalvani:
> But.. the link would be correct.
> 
> And right now, the links are broken, how is htmlproofer passing? Is it checking image links?

They are just broken when we deploy, locally they all work because the root folder is server at `/` instead of `/ping-viewer`.

Williangalvani:
It will probably work with absolute paths if we rename the "site" folder to "ping-viewer"

Williangalvani:
[this](https://github.com/bluerobotics/ping-viewer/compare/docs...Williangalvani:absolutePaths?expand=1) is the alternative

jaxxzer:
> [this](https://github.com/bluerobotics/ping-viewer/compare/docs...Williangalvani:absolutePaths?expand=1) is the alternative

Open the pr please.

Mistake here: https://github.com/bluerobotics/ping-viewer/compare/docs...Williangalvani:absolutePaths?expand=1#diff-41dee870c880e4d661d37a8ac2223342R1

Williangalvani:
replaced by https://github.com/bluerobotics/ping-viewer/pull/494

PR #494, opened 2019-04-05
Fix docs to work with absolute path

Williangalvani:


jaxxzer:
@Williangalvani have you tried building/hosting locally?

Williangalvani:
> @Williangalvani have you tried building/hosting locally?

Yeah, worked well here

PR #499, opened 2019-04-11
[1] Start Ping360 support integration with new protocol structure

patrickelectric:
Needs: https://github.com/bluerobotics/ping-protocol/pull/132
This update ping-submodule to use master
Also allows device manager development
Needs Ping 3.27 firmware release.
^ https://github.com/bluerobotics/ping-firmware-private/releases/tag/3.27rc

### DIFF
--- a/src/sensor/protocoldetector.cpp
+++ b/src/sensor/protocoldetector.cpp
@@ -7,8 +7,7 @@
 #include <QSerialPortInfo>
 #include <QUdpSocket>
 
-#include "pingmessage/pingmessage.h"
-#include "pingmessage/pingmessage_ping1D.h"
+#include "pingmessage/ping_ping1d.h"
 #include "protocoldetector.h"
 
 Q_LOGGING_CATEGORY(PING_PROTOCOL_PROTOCOLDETECTOR, "ping.protocol.protocoldetector")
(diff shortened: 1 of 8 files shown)


patrickelectric:
@jaxxzer @Williangalvani ^

src/sensor/protocoldetector.cpp
```diff
@@ -147,8 +146,8 @@ bool ProtocolDetector::checkSerial(LinkConfiguration& linkConf)
 bool ProtocolDetector::checkUdp(LinkConfiguration& linkConf)
 {
     // To find a ping, we this message on a link, then wait for a reply
-    ping_msg_ping1D_empty req;
-    req.set_id(Ping1DNamespace::Firmware_version);
+    ping_message_general_request req;
```
jaxxzer: Are we ready to make this change?

patrickelectric: Needs 3.27 to be release first.

jaxxzer:
@patrickelectric we should format the request according to the firmware version. We can remove the retro-support after ~6 months.

src/sensor/protocoldetector.cpp
jaxxzer: @patrickelectric we should format the request according to the firmware version. We can remove the retro-support after ~6 months.

patrickelectric: `general_request` is ok to use in `ProtocolDetector`, the only problem that is know after Ping1D release is while using `general_request` with `profile`. This patch is already compatible with old firmware versions.

PR #501, opened 2019-04-12
Restart application after reset settings

### LINKED ISSUE #440: Restart program automatically after reset settings.
The message we give out now says that we will do this, but it doesn't happen. Maybe we should change it in the meantime.

patrickelectric:
Tested in:
- [x] Linux
- [x] Windows
- [x] Mac

Fix #440

### DIFF
--- a/src/util/util.cpp
+++ b/src/util/util.cpp
@@ -72,6 +72,17 @@ void Util::update(QtCharts::QAbstractSeries* series, const QVector<double>& poin
     xySeries->replace(realPoints);
 }
 
+void Util::restartApplication()
+{
+    QCoreApplication::quit();
+    const QStringList arguments = QCoreApplication::arguments();
+    if(arguments.isEmpty()) {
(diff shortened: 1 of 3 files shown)


patrickelectric:
Tested and working, downloads [available here](https://github.com/bluerobotics/ping-viewer/releases/tag/continuous-restart).

jaxxzer:
Thanks @patrickelectric. Do you have any reference?

patrickelectric:
@jaxxzer reference for what exactly ?

src/util/util.cpp
```diff
@@ -72,6 +72,17 @@ void Util::update(QtCharts::QAbstractSeries* series, const QVector<double>& poin
     xySeries->replace(realPoints);
 }
 
+void Util::restartApplication()
```
jaxxzer: I mean did you use some example to create this, or did you create the solution yourself?

patrickelectric: Added.

PR #509, opened 2019-05-31
WIP: Rework waterfall to create a base for ping360

patrickelectric:
This is a rework of Waterfall to provide WaterfallPlot(Ping1D) and allow a common code for RadialPlot (Ping360)


src/waterfall/waterfall.h
```diff
@@ -18,42 +18,69 @@ class Waterfall : public QQuickPaintedItem
...
+     * @brief Construct a new Waterfall object
      *
+     * @param parent
      */
-    Q_INVOKABLE void clear();
+    Waterfall(QQuickItem *parent = nullptr);
 
     /**
-     * @brief Return waterfall image
+     * @brief This is used by the qml pain event
```
jaxxzer: * @brief This is used by the qml **paint** event

src/waterfall/waterfall.h
```diff
@@ -18,42 +18,69 @@ class Waterfall : public QQuickPaintedItem
...
+     *  This paint the waterfall in qml
      *
-     * @return QImage
+     * @param painer
```
jaxxzer: **painter**

src/waterfall/waterfall.h
```diff
@@ -18,42 +18,69 @@ class Waterfall : public QQuickPaintedItem
...
      */
-    QImage image() {return _image;}
-    Q_PROPERTY(QImage image READ image WRITE setImage NOTIFY imageChanged)
+    virtual void paint(QPainter *painer) = 0;
```
jaxxzer: **painter**

src/waterfall/waterfallplot.h
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * @brief Get signal confidence from mouse position column
+     *
+     * @return float
+     */
+    float mouseColumnConfidence() {return _mouseColumnConfidence;}
+    Q_PROPERTY(float mouseColumnConfidence READ mouseColumnConfidence NOTIFY mouseColumnConfidenceChanged)
+
+    /**
+     * @brief Get signal deepth from mouse position column
```
jaxxzer: **depth**

patrickelectric:
@jaxxzer 	fixed.

qml/Ping360Visualizer.qml
```diff
@@ -0,0 +1,120 @@
...
+            Layout.preferredWidth: 100
+            Layout.minimumWidth: 75
+        }
+
+        Settings {
+            property alias chartWidth: chart.width
+        }
+    }
+
+    ValueReadout {
```
jaxxzer: This ought to be removed, right? We don't have any 'depth' or 'confidence'.

src/waterfall/polarplot.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+#include <QtConcurrent>
+#include <QPainter>
+#include <QtMath>
+#include <QVector>
+
+PING_LOGGING_CATEGORY(polarplot, "ping.polarplot")
+
+// Number of samples to display
+uint16_t PolarPlot::_resolution = 400;
```
jaxxzer:
```suggestion
// number of pie slices in a full circle
uint16_t PolarPlot::_angularResolution = 400;
```

src/waterfall/polarplot.cpp
```diff
@@ -0,0 +1,114 @@
...
+    // This ring vector will store variables of the last n samples for user access
+    static const float rad2grad = 200/M_PI;
+    _DCRing[int(actualAngle*rad2grad) % 400] = {initPoint, length, confidence, distance};
+
+    for(int i = 1; i < center.x(); i++) {
+        pointColor = valueToRGB(points[i%200]);
+        step = ceil(i*2*d2r)*1.15;
+        for(float u = -0.25; u <= 0.25; u += 1/step) {
+            angleStep = u*d2r+actualAngle - M_PI/2;
+            _image.setPixelColor(center.x() + i*cos(angleStep), center.y() + i*sin(angleStep), pointColor);
```
jaxxzer: A consideration for optimizing: all of the trigonometry can be done at compile time. In a previous implementation, I identified each possible 'cell' to color as an individual QPath, and all of the QPaths were stored in a two dimensional lookup table accessed by (angle, range):
https://github.com/bluerobotics/ping-viewer/commit/228d66c88854ccd1e8bfa409af3a48cd431a4c80

I'm not sure which implementation is better for supporting dynamic # samples though (ex rectangle container scan).

qml/Ping360Visualizer.qml
```diff
@@ -0,0 +1,120 @@
...
+import WaterfallPlot 1.0
+import PolarPlot 1.0
+
+import SettingsManager 1.0
+import StyleManager 1.0
+
+Item {
+    id: visualizer
+    property alias waterfallItem: waterfall
+    property var protocol
```
jaxxzer: is it used?

qml/Ping360Visualizer.qml
```diff
@@ -0,0 +1,120 @@
...
+            Layout.fillHeight: true
+            Layout.fillWidth: true
+
+            PolarPlot {
+                id: waterfall
+                height: Math.min(parent.height, parent.width)
+                width: height
+                anchors.centerIn: parent
+
+                Shape {
```
jaxxzer: TODO: this should be shaped like a 'pie slice' instead of a rectangle, and the angular resolution should be quantized to match the sensor (400 steps).

qml/Ping360Visualizer.qml
patrickelectric: I believe that we can use it for tests and remove before merging with master.

qml/Chart.qml
```diff
@@ -97,8 +97,14 @@ Item {
 
         function draw(points, depth, initPost) {
             if (points) {
-                Util.update(serie, points, initPost, depth, minDepthToDraw, maxDepthToDraw, 1)
-                Util.update(serieInv, points, initPost, depth, minDepthToDraw, maxDepthToDraw, -1)
+                // If there is no user configuration, we set the max and min automatically
+                if(maxDepthToDraw == 0 && minDepthToDraw == 0) {
```
jaxxzer: I think this might be a little obscure place to put this logic. maybe we should TODO to refactor for readability, multiple sensors, etc.

src/waterfall/polarplot.h
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * @brief Draw a list of points in the waterfall
+     *
+     * @param points
+     * @param confidence
+     * @param initPoint
+     * @param length
+     * @param distance
+     */
+    Q_INVOKABLE void draw(const QVector<double>& points, float confidence = 0, float initPoint = 0, float length = 50,
```
jaxxzer: @patrickelectric you intend to rework the parameters here correct? (perhaps add a todo in those cases)

src/waterfall/polarplot.cpp
jaxxzer: we can leave it unresolved for posterity (:

jaxxzer: https://github.com/bluerobotics/ping-viewer/issues/515 I just found it, thank you!

jaxxzer:
A good move @patrickelectric . We need to be thoughtful of the architecture as we proceed to ensure it will enable all of our intended use cases.

src/waterfall/polarplot.h
patrickelectric: Yes

patrickelectric:
> 1. radar is polar (BUT we should also be allowed to plot the incoming data using the waterfall visualizer format for sidescan applications)

Yes, that's why both plots use the same base class, to allow this kind of use. 



> 3\. radar does not need to redraw entire image to plot a single sample (this is the case now correct?)

Yes, only 1 gradian per sample.


> 4\. There is no capability for confidence or distance measurement (though, it's possible in the future, we could do virtual or similar)

Yes, we can remove it before the release if no such capability is implemented. 

For all the other requests, please check the ping360 todo list here:
https://github.com/bluerobotics/ping-viewer/issues/515

kklemens:
> radar minimum distance is always the same (0) (I don't see any utility in plotting a donut)

Correct, it should always be zero. The black "gap" is actually the water column. Slant range to first return from seabed. (Putting my thoughts together for an operational guide on how to explain all these things)
Radially Scanning Above

PR #516, opened 2019-06-05
Create abstract ping sensor class to help with ping360 integration and UI modularity  

patrickelectric:
This PR improves ping360 integration with a base class, a simple simulation and frontend capabilities.
https://0x0.st/zuO8.mp4


src/sensor/pingsensor.h
```diff
@@ -0,0 +1,210 @@
...
+     *
+     * @return uint8_t source id
+     */
+    uint8_t srcId() const { return _srcId; }
+    Q_PROPERTY(int srcId READ srcId NOTIFY srcIdUpdate)
+
+    /**
+     * @brief Return destiny ID
+     *
+     * @return uint8_t destiny id
```
jaxxzer: destination

src/sensor/pingsensor.h
```diff
@@ -0,0 +1,210 @@
...
+     */
+    uint16_t firmware_version_patch() const { return _firmware_version_patch; }
+    Q_PROPERTY(int firmware_version_patch READ firmware_version_patch NOTIFY firmwareVersionPatchUpdate)
+
+    /**
+     * @brief Return device type number
+     *
+     * @return uint8_t Device type number
+     */
+    uint8_t device_type() const { return _device_type; }
```
jaxxzer: device_revision: https://github.com/bluerobotics/ping-protocol#4-device_information

src/sensor/pingsensor.h
```diff
@@ -0,0 +1,210 @@
+#pragma once
+
+#include "sensor.h"
+
+/**
+ * @brief Abstract ping sensors
+ *
+ */
+class PingSensor : public Sensor
```
jaxxzer: want to add protocol version member?

src/devicemanager/devicemanager.cpp
```diff
@@ -30,9 +30,8 @@ void DeviceManager::append(const LinkConfiguration& linkConf, const QString& dev
 {
     for(int i{0}; i < _sensors[Connection].size(); i++) {
```
jaxxzer: `int i{0}` :eyes:

qml/Ping360Visualizer.qml
```diff
@@ -8,12 +8,24 @@ import Qt.labs.settings 1.0
 import WaterfallPlot 1.0
 import PolarPlot 1.0
 
+import DeviceManager 1.0
 import SettingsManager 1.0
 import StyleManager 1.0
 
 Item {
-    id: visualizer
+    id: root
```
jaxxzer: does this file/hunk belong in this commit?

src/sensor/ping.h
```diff
@@ -36,23 +36,7 @@ class Ping : public Sensor
...
-    Q_PROPERTY(int srcId READ srcId NOTIFY srcIdUpdate)
-
-    /**
-     * @brief Return destiny ID
-     *
-     * @return uint8_t destiny id
-     */
-    uint8_t dstId() { return _dstId; }
-    Q_PROPERTY(int dstId READ dstId NOTIFY dstIdUpdate)
+    void printSensorInformation() const final;
```
jaxxzer: You should perhaps add `override` to be explicit. @tcanabrava

src/waterfall/polarplot.cpp
```diff
@@ -70,14 +70,15 @@ void PolarPlot::draw(const QVector<double>& points, float confidence, float init
...
-    static const float rad2grad = 200/M_PI;
-    _DCRing[int(actualAngle*rad2grad) % 400] = {initPoint, length, confidence, distance};
+    static const float rad2grad = _angularResolution/(2*M_PI);
+    _DCRing[int(actualAngle*rad2grad) % _angularResolution] = {initPoint, length, confidence, distance};
 
-    for(int i = 1; i < center.x(); i++) {
-        pointColor = valueToRGB(points[i]);
+    const float linearFactor = points.size()/(float)center.x();
+    for(int i = 0; i < center.x(); i++) {
+        pointColor = valueToRGB(points[i*linearFactor]);
```
jaxxzer: indexing with a float is... strange.

qml/MainPage.qml
```diff
@@ -246,18 +246,8 @@ Item {
...
-
-            ping1DVisualizer.waterfallItem.grabToImage(function(result) {
-                print("Grab waterfall image callback")
-                print(FileManager.createFileName(FileManager.Pictures))
-                result.saveToFile(FileManager.createFileName(FileManager.Pictures))
-            })
-        } else if (event.key == Qt.Key_R) {
-            print("Clear waterfall")
-            ping1DVisualizer.waterfallItem.clear()
+        } else {
```
jaxxzer: I think this might be a move in the wrong direction.

It seems not the job of the visualizer to handle keyboard keys. An idea instead of doing the keyboard handling for each visualizer, maybe add grabtoimage and clear functions to the visualizer, and call those with the keyboard shortcut from main qml here.

src/sensor/pingsensor.h
patrickelectric: device_type follows the protocol, I'll consider your comment as a point that device_revision is missing. Will add a new property.

src/devicemanager/devicemanager.cpp
patrickelectric: This is not related to this patch, we can fix that later.

src/sensor/ping.h
patrickelectric: final implies in override after c++11 if I'm not mistaken. `override final`is a bit redundant.

src/sensor/pingsensor.h
```diff
@@ -0,0 +1,210 @@
...
+     */
+    uint8_t device_type() const { return _device_type; }
+    Q_PROPERTY(int device_type READ device_type NOTIFY deviceTypeUpdate)
+
+    /**
+     * @brief Return device model number
+     *
+     * @return uint8_t Device model number
+     */
+    uint8_t device_model() const { return _device_model; }
```
jaxxzer: The [comment](https://github.com/bluerobotics/ping-viewer/pull/516/files#r292068257) was intended to go here:

> device_revision: https://github.com/bluerobotics/ping-protocol#4-device_information

src/sensor/ping.h
jaxxzer: > final implies in override after c++11 if I'm not mistaken.

Correct. The suggestion is **to be explicit**.

> override final is a bit redundant.

Not really, they are independent specifiers with independent meaning. You have just created a situation where the compiler can make an implicit decision based on the language specification. This is not clear to everyone.

qml/MainPage.qml
patrickelectric: Fixed.

src/sensor/ping.h
patrickelectric: Done.

PR #517, opened 2019-06-12
Improve polar draw interface with axis and corrections

patrickelectric:
image

### DIFF
--- a/qml/Ping360Visualizer.qml
+++ b/qml/Ping360Visualizer.qml
@@ -1,9 +1,9 @@
 import QtGraphicalEffects 1.0
-import QtQuick 2.7
+import QtQuick 2.12
 import QtQuick.Controls 2.2
 import QtQuick.Controls 1.4 as QC1
 import QtQuick.Layouts 1.3
-import QtQuick.Shapes 1.0
+import QtQuick.Shapes 1.12
 import Qt.labs.settings 1.0
 import WaterfallPlot 1.0
 import PolarPlot 1.0
@@ -68,6 +68,55 @@ Item {
                             x: waterfall.width*Math.cos(angle)/2
                             y: waterfall.height*Math.sin(angle)/2
                         }
+                    }
+                }
+
+                // Spinner that shows the head angle
+                Shape {
+                    id: shapeSpinner
+                    opacity: 1
+                    anchors.fill: parent
+                    vendorExtensionsEnabled: false
+                    property var angle: 0
+                    property var centerX: waterfall.width/2
+                    property var centerY: waterfall.height/2
+                    property var radius: waterfall.width/2
+
+                    ShapePath {
+                        id: shapePathSpinner
+                        strokeWidth: -1
+
+                        fillGradient: LinearGradient {
+                            x1: x2*1.1
+                            y1: y2*1.5
+                            x2: shapeSpinner.radius
+                            y2: shapeSpinner.radius
+                            GradientStop { position: 1; color: "transparent" }
+                            GradientStop { position: 0.8; color: "#7bd82c" }
+                        }
+
+                        // Create spinner outside arc
+                        PathAngleArc {
+                            centerX: shapeSpinner.centerX
+                            centerY: shapeSpinner.centerY
+                            radiusX: shapeSpinner.radius
+                            radiusY: shapeSpinner.radius
+                            sweepAngle: 0.5
+                            startAngle: -sweepAngle
+                        }
+
+                        // Create line between arc and origin of the chart
+                        PathLine {
+                            x: shapeSpinner.centerX
+                            y: shapeSpinner.centerY
+                        }
+                    }
+
+                    transform: Rotation {
+                        origin.x: shapeSpinner.centerX
+                        origin.y: shapeSpinner.centerY
+                        axis { x: 0; y: 0; z: 1 }
+                        angle: shapeSpinner.angle - 90
                     }
                 }
 
--- a/qml/Ping360Visualizer.qml
+++ b/qml/Ping360Visualizer.qml
@@ -24,7 +24,7 @@ Item {
 
         onDataChanged: {
             // Move from mm to m
-            root.draw(ping.data, 100, 0, 100, 0)
+            root.draw(ping.data, ping.angle, 0, 100)
         }
     }
 
@@ -34,8 +34,8 @@ Item {
         }
     }
 
-    function draw(points, confidence, initialPoint, length, distance) {
-        waterfall.draw(points, confidence, initialPoint, length, distance)
+    function draw(points, angle, initialPoint, length) {
+        waterfall.draw(points, angle, initialPoint, length)
         chart.draw(points, length + initialPoint, initialPoint)
     }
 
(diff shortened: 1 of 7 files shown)


src/waterfall/polarplot.cpp
```diff
@@ -59,31 +59,33 @@ void PolarPlot::setImage(const QImage &image)
...
     _DCRing[int(actualAngle*rad2grad) % _angularResolution] = {initPoint, length, confidence, distance};
+    static float angleResolution = g2d/2;
+    float actualAngle = g2d*angle*d2r;
 
     const float linearFactor = points.size()/(float)center.x();
     for(int i = 0; i < center.x(); i++) {
         pointColor = valueToRGB(points[static_cast<int>(i*linearFactor)]);
         step = ceil(i*2*d2r)*1.15;
-        for(float u = -0.25; u <= 0.25; u += 1/step) {
+        for(float u = -angleResolution; u <= angleResolution; u += 1/step) {
```
jaxxzer: Beware: https://stackoverflow.com/a/16595884

src/waterfall/polarplot.h
```diff
@@ -102,22 +102,9 @@ class PolarPlot : public Waterfall
...
-     *
-     */
-    struct DCPack {
-        float initialDepth;
-        float length;
-        float confidence;
-        float distance;
-    };
-
-    QVector<DCPack> _DCRing;
```
jaxxzer: We will need to scale previous data if the range changes. We may be able to accomplish this without a vector of past data by performing a transform on the current image. We can also do a redraw and erase previous data for the first solution because range will not change automatically/frequently.

qml/Ping360Visualizer.qml
```diff
@@ -86,9 +86,9 @@ Item {
                         id: mouseConfidenceText
                         x: mouseReadout.width - width
                         y: mouseReadout.height*4/5
-                        text: transformValue(waterfall.mouseSampleConfidence) + "%"
-                        visible: typeof(waterfall.mouseSampleConfidence) == "number"
-                        color: confidenceToColor(waterfall.mouseSampleConfidence)
+                        text: transformValue(waterfall.mouseSampleAngle) + "º"
+                        visible: typeof(waterfall.mouseSampleAngle) == "number"
+                        color: "green"
```
jaxxzer: The contrast is not very good. It might be more useful to give the text a static position in the corner (not obscuring data).

qml/PolarGrid.qml
```diff
@@ -0,0 +1,107 @@
...
+    width: 400
+    height: width
+
+    Shape {
+        opacity: 0.5
+        anchors.fill: parent
+        vendorExtensionsEnabled: false
+
+        // polar axis
+        ShapePath {
```
jaxxzer: We will require the auto increment adjustment / dynamic scaling like `DepthAxis`: https://github.com/bluerobotics/ping-viewer/blob/master/qml/DepthAxis.qml#L16

qml/Ping360Visualizer.qml
```diff
@@ -71,6 +71,51 @@ Item {
...
+                    ShapePath {
+                        id: shapePathSpinner
+                        strokeWidth: -1
+                        property var centerX: waterfall.width/2
+                        property var centerY: waterfall.height/2
+
+                        fillGradient: LinearGradient {
+                            x1: x2*1.1
+                            y1: y2*1.5
+                            x2: waterfall.width/2
```
jaxxzer: `shapePathSpinner.centerX`?

qml/Ping360Visualizer.qml
```diff
@@ -71,6 +71,51 @@ Item {
...
+                            startAngle: -sweepAngle
+                        }
+                        PathLine {
+                            x: shapePathSpinner.centerX
+                            y: shapePathSpinner.centerY
+                        }
+                    }
+
+                    transform: Rotation {
+                        origin.x: waterfall.width/2
```
jaxxzer: `shapePathSpinner.centerX`?

src/waterfall/polarplot.cpp
```diff
@@ -45,7 +45,7 @@ void PolarPlot::paint(QPainter *painter)
 
     // http://blog.qt.io/blog/2006/05/13/fast-transformed-pixmapimage-drawing/
     pix = QPixmap::fromImage(_image, Qt::NoFormatConversion);
-    _painter->drawPixmap(_painter->viewport(), pix, QRect(0, 0, _image.width(), _image.height()));
+    _painter->drawPixmap(QRect(0, 0, width(), height()), pix, QRect(0, 0, _image.width(), _image.height()));
```
jaxxzer: I recommend that you change the commit message to something like `polarplot: fix for high resolution displays (mac)'

qml/PolarGrid.qml
```diff
@@ -0,0 +1,107 @@
+import QtQuick 2.12
+import QtQuick.Shapes 1.12
+
+Item {
+    id: root
+    width: 400
+    height: width
+
+    Shape {
+        opacity: 0.5
```
jaxxzer: I prefer 100% opacity and 1~2 pixels wide.

qml/Ping360Visualizer.qml
```diff
@@ -71,6 +71,51 @@ Item {
...
+                Shape {
+                    id: shapeSpinner
+                    opacity: 1
+                    anchors.fill: parent
+                    vendorExtensionsEnabled: false
+                    property var angle: 0
+
+                    ShapePath {
+                        id: shapePathSpinner
+                        strokeWidth: -1
```
jaxxzer: Why is this -1?

qml/Ping360Visualizer.qml
```diff
@@ -71,6 +71,51 @@ Item {
                     }
                 }
 
+                // Spinner that shows the head angle
+                Shape {
```
jaxxzer: It is not immediately clear what each child's purpose is. A brief comment will help :)

src/waterfall/polarplot.h
patrickelectric: Right now we don't have confidence information, we can do a better approach removing/changing the structure for ping360. Waiting hardware integration to add distance/depth indirect calculation.

qml/Ping360Visualizer.qml
patrickelectric: Remove the stroke from the path.

qml/PolarGrid.qml
patrickelectric: Check https://github.com/bluerobotics/ping-viewer/pull/517#discussion_r293353491.

qml/Ping360Visualizer.qml
patrickelectric: This is not related to the center point but with the radius of the ring. 
Added radius variable to be more clear.

qml/Ping360Visualizer.qml
patrickelectric: done.

src/waterfall/polarplot.cpp
patrickelectric: This will result in a more complex logic and not clear math... But ok.

patrickelectric:
@jaxxzer 	PR updated with the requested modifications. Merge at will.
image

PR #519, opened 2019-06-14
PolarGrid: Add angular grid

patrickelectric:

### DIFF
--- a/qml/PolarGrid.qml
+++ b/qml/PolarGrid.qml
@@ -103,4 +103,29 @@ Item {
             }
         }
     }
+
+    Canvas {
+        anchors.fill: parent
+        opacity: 0.8
+        property var angle: 45
+
+        onPaint: {
+            var centerX = root.width/2
+            var centerY = root.height/2
+            var radius = root.height/2 - shapePathPolar.strokeWidth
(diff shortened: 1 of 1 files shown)


qml/PolarGrid.qml
```diff
@@ -103,4 +103,29 @@ Item {
             }
         }
     }
+
+    Canvas {
+        anchors.fill: parent
+        opacity: 0.8
+        property var angle: 45
+
+        onPaint: {
```
jaxxzer: Does this redraw every time we paint a new profile? If so, is that necessary? Can it just belong on another layer/item?

patrickelectric: No. What is necessary ? Why it should be in another item ? This is the polar grid, it should deal with grid stuff.

jaxxzer: My question is:

Does **this item** get drawn/painted once at initialization, or repeatedly every time the visualizer updates the sensor data?

jaxxzer: (when is onPaint called)

patrickelectric: As said before. No.
Paint is only called when the region needs to be rendered, like resize events.

PR #521, opened 2019-06-17
Random updates and corrections around ping360 integration 

patrickelectric:
- Add dynamic background
image
- Add controlPanel
image
- Improve PingTextField


src/sensor/ping.cpp
```diff
@@ -552,6 +552,17 @@ void Ping::resetSensorLocalVariables()
...
 }
 
+QQuickItem* Ping::controlPanel(QObject *parent)
+{
+    QQmlEngine *engine = qmlEngine(parent);
+    if(!engine) {
+        qCDebug(PING_PROTOCOL_PING) << "No qml engine to load visualization.";
+        return nullptr;
+    }
+    QQmlComponent *component = new QQmlComponent(engine, QUrl("qrc:/Ping1DControlPanel.qml"));
```
jaxxzer: How is this deleted? Does engine own it and do the clean up?

qml/Ping360ControlPanel.qml
```diff
@@ -0,0 +1,109 @@
...
+                            ping.sample_period = parseInt(text)*40
+                        }
+                    }
+                }
+
+                RowLayout {
+                     PingTextField {
+                        id: transmitFrequency
+                        title: "Transmit Frequency (kHz):"
+                        text: ping.transmit_frequency
```
jaxxzer: `[15:42:11.200] default[Warning]: Ping360ControlPanel.qml(94) qrc:/Ping360ControlPanel.qml:94:25: Unable to assign [undefined] to QString`

src/sensor/ping.cpp
patrickelectric: Added smart pointer to remove any possible memory leak, the parent should take of it, but better safe than sorry.

qml/Ping360ControlPanel.qml
patrickelectric: Fixed.

PR #522, opened 2019-06-18
Update submodule to ping-cpp

jaxxzer:
Before merging this:
- merge ping-protocol pr: https://github.com/bluerobotics/ping-protocol/pull/156
- update ping-protocol submodule in ping-cpp and force push (so that we use a commit from ping-protocol master)
- update the ping-cpp submodule here (so that we use a commit from ping-cpp master)

### DIFF
--- a/pingviewer.pro
+++ b/pingviewer.pro
@@ -22,8 +22,8 @@ RESOURCES += \
     QMAKE_LFLAGS += -fopenmp
 }
 
-include(lib/ping-protocol-cpp/ping.pri)
 include(lib/maddy/maddy.pri)
+include(lib/ping-cpp/ping-cpp.pri)
 
 CONFIG(debug, debug|release) {
     message("Debug Build !")
--- a/qml/Ping1DControlPanel.qml
+++ b/qml/Ping1DControlPanel.qml
@@ -37,7 +37,7 @@ Item {
                     PingButton {
                         text: "Emit Ping"
                         //requestEchosounderProfile
-                        onClicked: ping.request(Ping1DNamespace.Profile)
+                        onClicked: ping.emitPing()
                     }
 
                     Slider {
--- a/qml/Ping1DControlPanel.qml
+++ b/qml/Ping1DControlPanel.qml
@@ -6,7 +6,6 @@ import QtQuick.Controls.Material 2.2
 import QtQuick.Layouts 1.3
 
 import DeviceManager 1.0
-import Ping1DNamespace 1.0
 import SettingsManager 1.0
 import StyleManager 1.0
 
--- a/src/link/simulationlink.h
+++ b/src/link/simulationlink.h
@@ -1,7 +1,7 @@
 #pragma once
 
 #include "abstractlink.h"
-#include "parsers/parser.h"
+#include "parser-ping.h"
 
 /**
  * @brief Simulation connection class
@@ -24,7 +24,4 @@ public:
      * @return false
      */
     bool isWritable() final { return false; };
-
-private:
-    Parser _parser;
 };
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -25,9 +25,6 @@
 #include "util.h"
 #include "waterfallplot.h"
 
-// Register message enums to qml
-#include "pingmessage/ping_ping1d.h"
-
 Q_DECLARE_LOGGING_CATEGORY(mainCategory)
 
 PING_LOGGING_CATEGORY(mainCategory, "ping.main")
@@ -64,15 +61,6 @@ int main(int argc, char *argv[])
     qmlRegisterType<Ping360>("Ping360", 1, 0, "Ping360");
     qmlRegisterType<PolarPlot>("PolarPlot", 1, 0, "PolarPlot");
     qmlRegisterType<WaterfallPlot>("WaterfallPlot", 1, 0, "WaterfallPlot");
-
-    // Uncreatable register
-    qmlRegisterUncreatableMetaObject(
-        Ping1DNamespace::staticMetaObject, "Ping1DNamespace", 1, 0, "Ping1DNamespace", "This is a enum."
-    );
-
-    qmlRegisterUncreatableMetaObject(
-        Ping360Namespace::staticMetaObject, "Ping360Namespace", 1, 0, "Ping360Namespace", "This is a enum."
-    );
 
     qmlRegisterUncreatableMetaObject(
         AbstractLinkNamespace::staticMetaObject,
--- a/src/sensor/parser-ping.cpp
+++ b/src/sensor/parser-ping.cpp
@@ -1,0 +1,34 @@
+#include "parser-ping.h"
+
+void PingParserExt::clearBuffer()
+{
+    _parser.reset();
+}
+
+void PingParserExt::parseBuffer(const QByteArray& data)
+{
+    for(int i = 0; i < data.length(); i++) {
+        PingParser::ParseState state = _parser.parseByte(data.at(i));
+        if (state == PingParser::ParseState::NEW_MESSAGE) {
+            parsed++;
+            _rxMessage = _parser.rxMessage;
+            emit newMessage(_rxMessage);
+        } else if (state == PingParser::ParseState::ERROR) {
+            errors++;
+            emit parseError();
+        }
+    }
+}
+
+Parser::ParserState PingParserExt::parseByte(const char byte)
+{
+    PingParser::ParseState state = _parser.parseByte(byte);
+    if (state == PingParser::ParseState::NEW_MESSAGE) {
+        _rxMessage = _parser.rxMessage;
+        return Parser::ParserState::NEW_MESSAGE;
+    } else if (state == PingParser::ERROR) {
+        return Parser::ParserState::ERROR;
+    }
+
+    return Parser::ParserState::PARSING;
+}
--- a/src/sensor/parser-ping.h
+++ b/src/sensor/parser-ping.h
@@ -1,0 +1,38 @@
+#pragma once
+
+#include "parser.h"
+#include "ping-parser.h"
+
+/**
+ * @brief The PingParserExt class wraps the PingParser class from the ping-protocol submodule
+ * and Extends it with signalling in order to subclass our Parser class
+ */
+class PingParserExt : public Parser
+{
+public:
+    /**
+     * @brief Any messages parsed must be shorter than the buffer length
+     */
+    PingParserExt() : _parser(10240) {}
+
+    /**
+     * @brief clear parse state
+     */
+    void clearBuffer() override final;
+
+    /**
+     * @brief asynchronous use, Child classes should signal when something happens ie. 'emit newMessage(Message m)'
+     * @param data the next sequence of bytes in the serial stream being parsed
+     */
+    void parseBuffer(const QByteArray& data) override final;
+
+    /**
+     * @brief synchronous use, Child should return flags indicating incremental parse result/status
+     * @param byte the next byte in serial stream being parsed
+     * @return the current state of the parser
+     */
+    ParserState parseByte(const char byte) override final;
+
+private:
+    PingParser _parser;
+};
--- a/src/sensor/parser.h
+++ b/src/sensor/parser.h
@@ -1,0 +1,59 @@
+#pragma once
+
+#include <QObject>
+#include "ping-message.h"
+
+/**
+ * This class digests data and notifies owner when something interesting happens
+ */
+class Parser : public QObject
+{
+    Q_OBJECT
+public:
+    Parser() = default;
+    ~Parser() = default;
+
+    /**
+     * @brief parser states
+     */
+    enum ParserState {
+        ERROR, // got some unexpected data
+        PARSING, // got some expected data, feed me more
+        NEW_MESSAGE // got a new packet
+    };
+
+    uint32_t parsed = 0; // number of messages/packets successfully parsed
+    uint32_t errors = 0; // number of parse errors
+
+    /**
+     * @brief clear parse state
+     */
+    virtual void clearBuffer() = 0;
+
+    /**
+     * @brief asynchronous use, Child classes should signal when something happens ie. 'emit newMessage(Message m)'
+     * @param data the next sequence of bytes in the serial stream being parsed
+     */
+    virtual void parseBuffer(const QByteArray& data) = 0;
+
+    /**
+     * @brief synchronous use, Child should return flags indicating incremental parse result/status
+     * @param byte the next byte in serial stream being parsed
+     * @return the current state of the parser
+     */
+    virtual ParserState parseByte(const char byte) = 0;
+
+    /**
+     * @brief Access the last message that was parsed for synchronous use
+     *
+     * @return the last message that was parsed
+     */
+    ping_message rxMessage() const { return _rxMessage; }
+
+signals:
+    void newMessage(const ping_message& msg);
+    void parseError();
+
+protected:
+    ping_message _rxMessage;
+};
--- a/src/sensor/ping.h
+++ b/src/sensor/ping.h
@@ -6,9 +6,9 @@
 #include <QSharedPointer>
 #include <QTimer>
 
-#include "parsers/parser.h"
-#include "parsers/parser_ping.h"
-#include "pingmessage/ping_ping1d.h"
+#include <ping-message.h>
+#include <ping-message-common.h>
+#include <ping-message-ping1d.h>
 #include "protocoldetector.h"
 #include "pingsensor.h"
 
@@ -53,11 +53,11 @@ public:
      */
     void pingEnable(bool enabled)
     {
-        ping1D_set_ping_enable m;
+        ping1d_set_ping_enable m;
         m.set_ping_enabled(enabled);
         m.updateChecksum();
         writeMessage(m);
-        request(Ping1DNamespace::PingEnable);
+        request(Ping1dId::PING_ENABLE);
     }
 
     Q_PROPERTY(bool pingEnable READ pingEnable WRITE pingEnable NOTIFY pingEnableUpdate)
@@ -108,12 +108,12 @@ public:
      */
     void set_start_mm(int start_mm)
     {
-        ping1D_set_range m;
+        ping1d_set_range m;
         m.set_scan_start(start_mm);
         m.set_scan_length(_scan_length);
         m.updateChecksum();
         writeMessage(m);
-        request(Ping1DNamespace::Range);
+        request(Ping1dId::RANGE);
     }
     Q_PROPERTY(int start_mm READ start_mm WRITE set_start_mm NOTIFY scanStartUpdate)
 
@@ -131,12 +131,12 @@ public:
      */
     void set_length_mm(int length_mm)
     {
-        ping1D_set_range m;
+        ping1d_set_range m;
         m.set_scan_start(_scan_start);
         m.set_scan_length(length_mm);
         m.updateChecksum();
         writeMessage(m);
-        request(Ping1DNamespace::Range);
+        request(Ping1dId::RANGE);
     }
     Q_PROPERTY(int length_mm READ length_mm WRITE set_length_mm NOTIFY scanLengthUpdate)
 
@@ -154,11 +154,11 @@ public:
      */
     void set_gain_setting(int gain_setting)
     {
-        ping1D_set_gain_setting m;
+        ping1d_set_gain_setting m;
         m.set_gain_setting(gain_setting);
         m.updateChecksum();
         writeMessage(m);
-        request(Ping1DNamespace::GainSetting);
+        request(Ping1dId::GAIN_SETTING);
     }
     Q_PROPERTY(int gain_setting READ gain_setting WRITE set_gain_setting NOTIFY gainSettingUpdate)
 
@@ -185,23 +185,31 @@ public:
      */
     void set_mode_auto(bool mode_auto)
     {
-        ping1D_set_mode_auto m;
+        ping1d_set_mode_auto m;
         m.set_mode_auto(mode_auto);
         m.updateChecksum();
         writeMessage(m);
-        request(Ping1DNamespace::ModeAuto);
+        request(Ping1dId::MODE_AUTO);
     }
     Q_PROPERTY(bool mode_auto READ mode_auto WRITE set_mode_auto NOTIFY modeAutoUpdate)
+
+    /**
+     * @brief These are ids that the sensor can continuously output.
+     * The sensor only supports continuous profile output
+     */
+    enum ContinuousId : uint16_t {
+        PROFILE = Ping1dId::PROFILE
+    };
 
     /**
      * @brief Do continuous start
      *
      * @param id
      */
-    void do_continuous_start(Ping1DNamespace::ping1D_id id)
+    void do_continuous_start(ContinuousId id)
     {
-        ping1D_continuous_start m;
-        m.set_id(static_cast<int>(id));
+        ping1d_continuous_start m;
+        m.set_id(id);
         m.updateChecksum();
         writeMessage(m);
     }
@@ -211,9 +219,9 @@ public:
      *
      * @param id
      */
-    void do_continuous_stop(Ping1DNamespace::ping1D_id id)
+    void do_continuous_stop(ContinuousId id)
     {
-        ping1D_continuous_stop m;
+        ping1d_continuous_stop m;
         m.set_id(static_cast<int>(id));
         m.updateChecksum();
         writeMessage(m);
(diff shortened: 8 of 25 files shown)


patrickelectric:
Use device-manager, master branch is only for bug fixes right now related to ping1D.
It would be safer to move that in the actual wip branch.

qml/MainMenu.qml
```diff
@@ -6,7 +6,6 @@ import QtQuick.Controls.Material 2.2
 import QtQuick.Layouts 1.3
 import Qt.labs.settings 1.0
 
-import Ping1DNamespace 1.0
```
patrickelectric: This breaks "Emit Ping" button.

src/sensor/parser-ping.cpp
```diff
@@ -0,0 +1,20 @@
...
+        PingParser::ParseState state = parser_.parseByte(data.at(i));
+        if (state == PingParser::ParseState::NEW_MESSAGE) {
+            parsed++;
+            emit newMessage(ping_message(parser_.rxMessage));
+        } else if (state == PingParser::ParseState::ERROR) {
+            errors++;
+            emit parseError();
+        }
+    }
+}
```
patrickelectric: This function does not return any type of information related to the correct parser of a message, it should return a bool or a state condition to allow synchronous logic.

src/sensor/parser.h
```diff
@@ -0,0 +1,25 @@
...
+    Q_OBJECT
+public:
+    Parser() = default;
+    ~Parser() = default;
+
+    uint32_t parsed; // number of messages/packets successfully parsed
+    uint32_t errors; // number of parse errors
+
+    // asynchronous use, Child classes should signal when something happens ie. 'emit newMessage(Message m)'
+    virtual void parseBuffer(const QByteArray& data) { Q_UNUSED(data) }
```
patrickelectric: This should be pure virtual

src/sensor/parser.h
```diff
@@ -0,0 +1,25 @@
...
+    // synchronous use, Child should return flags indicating incremental parse result/status
+    virtual uint8_t parseByte(const char byte) { Q_UNUSED(byte) return 0; }
```
patrickelectric: This should be pure virtual

src/sensor/parser.h
```diff
@@ -0,0 +1,25 @@
...
+
```
patrickelectric: The class should return somehow a synchronous way to get the last parsed message.

src/sensor/parser.h
patrickelectric: Why is this function returning an `uint8_t` ?

src/sensor/parser.h
```diff
@@ -0,0 +1,25 @@
...
+signals:
+    void newMessage(ping_message msg);
```
patrickelectric: `const ping_message&`

pingviewer.pro
```diff
@@ -84,4 +84,4 @@ exists ($$_PRO_FILE_PWD_/.git) {
 message(" GIT_VERSION:      " $$GIT_VERSION)
 message(" GIT_VERSION_DATE: " $$GIT_VERSION_DATE)
 message(" GIT_TAG:          " $$GIT_TAG)
-message(" GIT_URL:          " $$GIT_URL)
\ No newline at end of file
+message(" GIT_URL:          " $$GIT_URL)
```
patrickelectric: Remove style changes.

src/link/pingsimulationlink.cpp
```diff
@@ -1,7 +1,9 @@
 #include <QtMath>
 
 #include "pingsimulationlink.h"
-#include "pingmessage/ping_ping1d.h"
+#include <ping-message.h>
```
patrickelectric: Why is this include here ?

src/link/pingsimulationlink.cpp
```diff
@@ -1,7 +1,9 @@
...
+#include <ping-message-common.h>
```
patrickelectric: The class is not using common messages

src/link/simulationlink.h
```diff
@@ -23,7 +23,7 @@ class SimulationLink : public AbstractLink
      * @return true
      * @return false
      */
-    bool isWritable() final { return false; };
+    bool isWritable() final { return false; }
```
patrickelectric: Avoid style changes

src/main.cpp
```diff
@@ -58,11 +55,6 @@ int main(int argc, char *argv[])
     qmlRegisterType<Ping>("Ping", 1, 0, "Ping");
     qmlRegisterType<Waterfall>("Waterfall", 1, 0, "Waterfall");
 
-    // Uncreatable register
-    qmlRegisterUncreatableMetaObject(
-        PingPing1DNamespace::staticMetaObject, "Ping1DNamespace", 1, 0, "Ping1DNamespace", "This is a enum."
```
patrickelectric: This is used in some qml files.

src/sensor/ping.h
```diff
@@ -253,9 +253,9 @@ class Ping : public Sensor
      *
      * @param id
      */
-    void do_continuous_start(PingPing1DNamespace::ping_ping1D_id id)
+    void do_continuous_start(uint16_t id)
```
patrickelectric: This is a high level class, it's not save to use uint16_t as parameter, use enum.
The enum is explicit and allows the compiler to detect problems.

src/sensor/ping.h
```diff
@@ -648,5 +648,5 @@ class Ping : public Sensor
         int waiting = 0;
     };
     friend QDebug operator<<(QDebug d, const Ping::messageStatus& other);
-    QHash<PingPing1DNamespace::ping_ping1D_id, messageStatus> requestedIds;
+    QHash<uint16_t, messageStatus> requestedIds;
```
patrickelectric: This is a high level class, it's not save to use uint16_t as key, use enum.
The enum is explicit and allows the compiler to detect problems.

src/test.cpp
```diff
@@ -171,4 +170,4 @@ void Test::waterfallGradient()
 
 }
 
-QTEST_MAIN(Test)
```
patrickelectric: Avoid style changes.

src/sensor/parser.h
```diff
@@ -0,0 +1,25 @@
...
+    void parseError();
+};
```
patrickelectric: This should contain a `clearBuffer()` function to remove any trace of old information from older tries for mult-sensor logic.

qml/MainMenu.qml
jaxxzer: @patrick I think instead of exporting pingmessage enums to qml, it's better to have a QINVOKABLE emitPing()

src/sensor/parser-ping.cpp
jaxxzer: This function is explicitly for asynchronous use only: https://github.com/bluerobotics/ping-viewer/blob/050fe06c7bba575b5a824c9406eba6c1aca2a476/src/sensor/parser.h#L17

Synchronous operation is only allowed with the parseByte() function  (as in arduino implementation).

jaxxzer: I've updated this class to allow synchronous operation according to your recommendation, but this function remains void return.

src/sensor/parser.h
jaxxzer: done

src/test.cpp
jaxxzer: Our guideline is a newline at the end of the file.

qml/MainMenu.qml
jaxxzer: I added emitPing(), it's less code for the whole project.

src/sensor/parser.h
jaxxzer: I have a problem when I attempt to make it pure virtual:
Screenshot_20190621_094559

src/link/simulationlink.h
jaxxzer: This is your preference (no semi colon). You want a new commit?

src/sensor/parser.h
jaxxzer: fixed

src/sensor/parser.h
jaxxzer: done

src/sensor/parser.h
jaxxzer: done

src/link/simulationlink.h
jaxxzer: I'll put the style changes in different commits.

src/main.cpp
jaxxzer: Not anymore!

src/sensor/parser.h
jaxxzer: done

src/link/pingsimulationlink.cpp
jaxxzer: removed

src/link/pingsimulationlink.cpp
jaxxzer: removed

src/sensor/ping.h
jaxxzer: a uint16_t is safe by definition - a message id is a uint16_t. To change the signature here will also require a single enum scope for all message ids (common, ping1d, ping360), which we do not have (we could make it if we require reserved ids for each sensor).

You can still gain the compiler/enum advantages: I recommend the caller make use of the the individual message id enum/namespaces.

src/sensor/ping.h
jaxxzer: done

src/link/simulationlink.h
jaxxzer: I'm dropping them for expediency.

src/test.cpp
jaxxzer: removed the change

src/sensor/parser.h
jaxxzer: I assume it's because we have Parser member in abstract link, it the functions need to be defined.

jaxxzer:
Update, all review has been addressed, and everything is working with the application locally. I am working now on adding class documentation for the parsers and fixing the ci fail.

jaxxzer:
The application works with ping1d simulation, but not with a real ping1d hardware. It looks like some problem with the request timer callback and/or the initializing sensor settings from previous session.

src/sensor/parser.h
patrickelectric: Why `parseBuffer` does not return state ? This will obligate us to create a loop and do what `parBuffer` is already doing.

src/sensor/parser.h
```diff
@@ -0,0 +1,39 @@
...
+    virtual ParserState parseByte(const char byte) { Q_UNUSED(byte) return ERROR; }
+
+    ping_message rxMessage() { return rxMessage_; }
+
+signals:
+    void newMessage(const ping_message& msg);
+    void parseError();
+
+protected:
+    ping_message rxMessage_;
```
patrickelectric: Please follow our code style:
https://github.com/bluerobotics/software-guidelines/blob/master/guidelines/style.md#code

src/sensor/parser.h
```diff
@@ -0,0 +1,39 @@
...
+
+    //TODO move to pure virtual
+    // clear parse state
+    virtual void clearBuffer() { }
+    // asynchronous use, Child classes should signal when something happens ie. 'emit newMessage(Message m)'
+    virtual void parseBuffer(const QByteArray& data) { Q_UNUSED(data) }
+    // synchronous use, Child should return flags indicating incremental parse result/status
+    virtual ParserState parseByte(const char byte) { Q_UNUSED(byte) return ERROR; }
+
+    ping_message rxMessage() { return rxMessage_; }
```
patrickelectric: const function.

src/sensor/ping.h
```diff
@@ -185,23 +185,28 @@ class Ping : public PingSensor
...
         m.updateChecksum();
         writeMessage(m);
-        request(Ping1DNamespace::ModeAuto);
+        request(Ping1dId::MODE_AUTO);
     }
     Q_PROPERTY(bool mode_auto READ mode_auto WRITE set_mode_auto NOTIFY modeAutoUpdate)
 
+    // sensor only supports continuous profile output
+    enum ContinuousId : uint16_t {
+        PROFILE = Ping1dId::PROFILE
```
patrickelectric: This doesn't look as a good approach, if `do_continuous_start` only supports one argument, it should not accept any argument at all.

src/sensor/ping.h
patrickelectric: From the creator of C++ language:

> enum class -- scoped and strongly typed enums
The enum classes ("new enums", "strong enums") address three problems with traditional C++ enumerations:
conventional enums implicitly convert to int, causing errors when someone does not want an enumeration to act as an integer.
conventional enums export their enumerators to the surrounding scope, causing name clashes.
the underlying type of an enum cannot be specified, causing confusion, compatibility problems, and makes forward declaration impossible.

http://www.stroustrup.com/C++11FAQ.html#enum

There is also the cpp guidelines:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#enum-enumerations

src/sensor/parser-ping.cpp
```diff
@@ -0,0 +1,34 @@
...
+void PingParserExt::clearBuffer()
+{
+    parser_.reset();
+}
+
+void PingParserExt::parseBuffer(const QByteArray& data)
+{
+    for(int i = 0; i < data.length(); i++) {
+        PingParser::ParseState state = parser_.parseByte(data.at(i));
+        if (state == PingParser::ParseState::NEW_MESSAGE) {
```
patrickelectric: Why ParseState is necessary ?

src/sensor/parser-ping.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include "parser.h"
+#include "ping-parser.h"
+
+class PingParserExt : public Parser
+{
+public:
+    PingParserExt() : parser_(2048) {}
```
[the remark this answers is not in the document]
jaxxzer: undefined behavior

src/sensor/parser-ping.h
```diff
@@ -0,0 +1,16 @@
...
+    void clearBuffer() override final;
+    void parseBuffer(const QByteArray& data) override final;
+    ParserState parseByte(const char byte) override final;
+
+private:
+    PingParser parser_;
```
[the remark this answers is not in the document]
jaxxzer: It will take some time, but I can change it

src/sensor/parser.h
jaxxzer: ok

src/sensor/parser.h
jaxxzer: I was reading some things that prefixing with underscores is not recommended for c/c++, it has some reserved or special meaning.

It will take some time, but I can change it.

src/sensor/ping.h
jaxxzer: The idea is to use it to request different messages, so it is the correct design for `do_continuous_start` to have a message id parameter. The sensor just doesn't support continuous output for `distance` messages yet. When it does, we will add it to the list here.

src/sensor/ping.h
jaxxzer: Thank you, I've read it and I understand the enum, the enum class, type safety and the advantages of using them. I still think it is the wrong idea to have an enum type for the parameter. (But I think it is the right idea for the caller to always use enum types for the arguments! ie `request(Ping1dId::DISTANCE)`. You can still write your code using enum types in every call.

Because you can, does not mean you should. Because it compiles, does not make it right. It is similar to our past conversation about if we should declare something const here:  https://github.com/bluerobotics/ping-protocol/pull/77

pingviewer.pro
```diff
@@ -22,8 +22,8 @@ RESOURCES += \
     QMAKE_LFLAGS += -fopenmp
 }
 
-include(lib/ping-protocol-cpp/ping.pri)
 include(lib/maddy/maddy.pri)
+include(lib/lib.pri)
```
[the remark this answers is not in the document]
jaxxzer: That seems unnecessary. What is the problem with this? Do you have some reference for the standard?

src/sensor/parser-ping.cpp
jaxxzer: It is to expose the current state of the parser. There are **three** possible states of interest: new message, parsing, and error.

src/sensor/parser-ping.cpp
jaxxzer: Because `parseBuffer` might parse more than one message in a single call. It was created to add async signalling to the class.

jaxxzer: We could make it bool return, but I don't see any use case to apply it to.

jaxxzer:
> The application works with ping1d simulation, but not with a real ping1d hardware.

update: my bug working with the real hardware is fixed

pingviewer.pro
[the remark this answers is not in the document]
jaxxzer: done, but I don't like the extra directory level to hold just a .pri.

src/sensor/parser-ping.h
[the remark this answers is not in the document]
jaxxzer: I added some documentation for the constructor

jaxxzer:
There are two remaining things here it will help me if you can take a look @patrickelectric, I can't see how my patch has affected these:
- the lost messages is initialized with a value of ~500k
- the control panel is not working from some undefined reference to ping in qml

lib/lib.pri
```diff
@@ -0,0 +1,5 @@
+MESSAGEPATH = $$PWD/ping-cpp/src/message
+
+HEADERS += $$MESSAGEPATH/*.h
+
+INCLUDEPATH += $$MESSAGEPATH
```
patrickelectric: This file is not necessary anymore.

src/sensor/parser-ping.cpp
patrickelectric: I'm saying that in some points you use `PingParser::NEW_MESSAGE`, `PingParser::ParseState::NEW_MESSAGE` and `Parser::ParserState::NEW_MESSAGE`.
There is two different classes and the enum name is not being used in some places resulting in 3 different enumeration uses. 
Is it possible to use the base class enum `Parser` with/without `ParserState` to help with readability and maintenance and avoiding the usage of a secondary kind ?

src/sensor/parser-ping.cpp
```diff
@@ -0,0 +1,34 @@
...
+}
+
+Parser::ParserState PingParserExt::parseByte(const char byte)
+{
+    uint8_t state = _parser.parseByte(byte);
+    if (state == PingParser::ParseState::NEW_MESSAGE) {
+        _rxMessage = _parser.rxMessage;
+        return Parser::ParserState::NEW_MESSAGE;
+    } else if (state == PingParser::ERROR) {
+        return Parser::ParserState::ERROR;
```
patrickelectric: Why we are not using the base class enum here ? What is the problem to use PingParser ?

src/sensor/parser-ping.h
```diff
@@ -0,0 +1,38 @@
...
+ * @brief The PingParserExt class wraps the PingParser class from the ping-protocol submodule
+ * and Extends it with signalling in order to subclass our Parser class
+ */
+class PingParserExt : public Parser
+{
+public:
+    /**
+     * @brief Any messages parsed must be shorter than the buffer length
+     */
+    PingParserExt() : _parser(2048) {}
```
patrickelectric: Take in mind that Ping360 `device_data` can use up to 2026 points (if the max profile length is 2000 points).
This is not a microcontroller, we can put 10240 here just to be safe, or if you want to be safer and avoid multiple page memories, you can use 4096.

src/sensor/ping.h
```diff
@@ -321,6 +329,11 @@ class Ping : public PingSensor
     Q_INVOKABLE void firmwareUpdate(QString fileUrl, bool sendPingGotoBootloader = true, int baud = 57600,
                                     bool verify = true);
 
+    /**
+     * @brief request a profile message from the device
+     */
+    Q_INVOKABLE void emitPing() { request(Ping1dId::PROFILE); }
```
patrickelectric: const function.

src/sensor/pingsensor.cpp
```diff
@@ -22,8 +24,9 @@ void PingSensor::request(int id)
     }
     qCDebug(PING_PROTOCOL_PINGSENSOR) << "Requesting:" << id;
 
-    ping_message_empty m;
-    m.set_id(id);
+    ping_message m(10);
```
patrickelectric: Why 10 is being used here ?
Should 10 be the default constructor ?

src/sensor/pingsensor.cpp
```diff
@@ -22,8 +24,9 @@ void PingSensor::request(int id)
...
+    m.set_payload_length(0);
```
patrickelectric: Wait, why is this necessary ? Why is not zero by default ?

qml/Ping1DControlPanel.qml
```diff
@@ -5,8 +5,6 @@ import QtQuick.Controls 2.2
 import QtQuick.Controls.Material 2.2
 import QtQuick.Layouts 1.3
 
-import DeviceManager 1.0
```
patrickelectric: This is necessary to get the connected device, otherwise the entire widget will not work.

src/sensor/parser.h
```diff
@@ -0,0 +1,60 @@
...
+     * @return the last message that was parsed
+     */
+    ping_message rxMessage() const { return _rxMessage; }
+
+signals:
+    void newMessage(const ping_message& msg);
+    void parseError();
+
+protected:
+    ping_message _rxMessage;
```
patrickelectric: Parser should be a generic class, no ? So why there is ping_message here ?
Should this be in PingParserExt ?

qml/Ping1DControlPanel.qml
jaxxzer: uuuuuff how did I miss this

src/sensor/pingsensor.cpp
jaxxzer: This is to construct an arbitrary pingmessage by hand.
Here we are making our 'empty' pingmessage with a message length of 10.

> Should 10 be the default constructor ?

Maybe. If we add 10 as default argument it will conflict with another constructor, i'm not sure if that one is used. We can/should definitely try to make the pingmessage class more robust to memory access violations.

src/sensor/parser-ping.cpp
jaxxzer: > Why we are not using the base class enum here ?

In our project, `Parser::ParseByte` returns the enum type `Parser::ParseState`: https://github.com/bluerobotics/ping-viewer/blob/protocol/src/sensor/parser.h#L19

In ping-cpp, `PingParser::parseByte` returns the enum type `PingParser::ParseState`: https://github.com/bluerobotics/ping-cpp/blob/deployment/src/message/ping-parser.h#L17

The base class Parser has no knowledge of the `PingParser` enum from ping-cpp. The PingParserExt method here does the conversion between the states.

jaxxzer: `Parser::ParseState::NEW_MESSAGE == PingParser::ParseState::NEW_MESSAGE` is not necessarily true, that's why the code is written like this.

src/sensor/pingsensor.cpp
jaxxzer: msgData is allocated on the heap and not zero initialized, nor is it initialized by this message constructor, because the constructor doesn't know anything about the message payload or it's length.

src/sensor/parser.h
jaxxzer: You are right... We could change the signal to `newMessage()`, then copy the message synchronously.

src/sensor/ping.h
jaxxzer: done, and now Sensor::request is const too as a requirement.

src/sensor/parser-ping.h
jaxxzer: done

lib/lib.pri
jaxxzer: done

src/sensor/parser.h
jaxxzer: This is a mistake from the past. I think we should leave it alone, and reevaluate it when we have the need for another parser/protocol.

jaxxzer:
@patrickelectric I think we are close.. let me know anything else to change.

src/sensor/pingsensor.cpp
jaxxzer: Our `ping_message_empty` constructor did the initialization for us manually in the past.

jaxxzer:
The only thing I don't know why it's changed, is the lost messages is initialized at this number every time:
Screenshot_20190626_235835

patrickelectric:
> The only thing I don't know why it's changed, is the lost messages is initialized at this number every time:
> Screenshot_20190626_235835

Fixed in https://github.com/bluerobotics/ping-viewer/pull/527

patrickelectric:
Can we merge this ?

PR #524, opened 2019-06-19
Update ping360 class

patrickelectric:


src/sensor/ping360.h
```diff
@@ -38,6 +38,27 @@ class Ping360 : public PingSensor
...
 
+    /**
+     * @brief Helper function to request profiles from sensor based in the actual position
+     *
+     * @param delta number of grads/steps from the actual position
+     */
+    Q_INVOKABLE void deltaStep(int delta)
+    {
+        ping360_transducer msg;
+        msg.set_mode(0);
```
Williangalvani: Shouldn't this be 1? https://github.com/bluerobotics/ping-protocol/blob/master/src/definitions/ping360.json#L101

Also, what about using an enum to make this and `set_transmit()` more readable?

patrickelectric: Indeed, but I did some reverse engineer over our working GUI for windows and the operation mode used by it is 0.
I got this specific and periodic array of:
`42 52 00 0e 0a 29 fe 01 00 01 00 92 00 2a 05 4c 02 e4 00 c8 01 00 04 91`
You can decode that following [our documentation here](https://docs.bluerobotics.com/ping-protocol/pingmessage-ping360):
BR                  = 42 52
payload_length      = 00 0e         = 13
message_id          = 0a 29         = 2601
src_device_id       = fe            = 254
dst_device_id       = 01            = 1
payload           
mode                = 00            = 0  
gain_setting        = 01            = 1
angle               = 00 92         = 5C
transmit_duration   = 00 2a         = 42
sample_period       = 05 4c         = 1356
transmit_frequency  = 02 e4         = 740
number_of_samples   = 00 c8         = 200
transmit            = 01            = 1
reserved            = 00            = 0
cs                  = 04 91

This PR follows this decoded array.

patrickelectric: > Also, what about using enums to make this and set_transmit() more readable?
YES! We need to wait @jaxxzer to finish the code generation to start doing that. But this will certainly be done.

Williangalvani: > Indeed, but I did some reverse engineer over our working GUI for windows and the operation mode used by it is 0.

damn, but have you tried with 1? Is the field just ignored by the FW for now?

patrickelectric: I can't see the difference, the sensor does not work in air. Will ask some people about it.

patrickelectric: The existent GUI interface appears to not provide such configuration.

patrickelectric: @Williangalvani 	It appears to be ignored by the firmware, I'll start with 1 since this is what we have in our documentation.

patrickelectric: Fixed.

Williangalvani:
I believe you can squash some things, at least 123a60921f304eb0f054997c4f772bc2d5d70a0c into e9b5b1534a22ec8ca527dcb722017ee6bef10f64, but nbd.
Other than that, it looks good!

patrickelectric:
Good catch, squashed and waiting for CI.

src/sensor/ping360.h
jaxxzer: > YES! We need to wait @jaxxzer to finish the code generation to start doing that. But this will certainly be done.

While you wait for me, you can create your own enums in your own scope (put them in version control in ping-viewer repository) in the intermediate term.

Mode here is undefined, there are no modes for the sensor. We should probably change the name of the field to `reserved`.

PR #525, opened 2019-06-20
Improve components provider calls and move to sensor base class

patrickelectric:


src/sensor/sensor.cpp
```diff
@@ -105,4 +108,60 @@ void Sensor::connectLinkLog(const LinkConfiguration& logConf)
     emit linkLogUpdate();
 }
 
+void Sensor::setControlPanel(const QUrl& url)
+{
+    if(!url.isValid()) {
+        qCWarning(PING_PROTOCOL_SENSOR) << "Invalid url:" << url;
```
Williangalvani: What about using Critical if these are invalid? You are unable to use the sensor properly without them, right?

src/sensor/sensor.cpp
```diff
@@ -105,4 +108,60 @@ void Sensor::connectLinkLog(const LinkConfiguration& logConf)
...
+        _sensorVisualizerUrl.setUrl({});
+        return;
+    }
+
+    _sensorVisualizerUrl = url;
+}
+
+QQuickItem* Sensor::controlPanel(QObject* parent)
+{
+    QQmlEngine* engine = qmlEngine(parent);
```
Williangalvani: Can we get comments or a reference here? Explaining what this does and how.

patrickelectric: This is a consequence of `QQmlComponent` constructor, since it's need an engine to create the component, and we are going to use the parent to use qml memory management to take care of that for us, we can extract the QQmlEngine from this parent (avoiding unnecessary methods and thread problems) to create our own component.

qml/NoControlPanel.qml
```diff
@@ -0,0 +1,23 @@
+import QtQml 2.2
+import QtQuick 2.7
+import QtQuick.Controls 2.2
+import QtQuick.Layouts 1.3
+
+GroupBox {
+    id: root
+    title: "No sensor"
+    label.x: width/2 - label.contentWidth/2
```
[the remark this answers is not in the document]
patrickelectric: AFAIK no.

jaxxzer:
nice :clap:

PR #527, opened 2019-06-27
Fix lost msgs property

patrickelectric:


patrickelectric:
@jaxxzer we were not initializing `_lostMessages` variable.
In ping, the variable was lost and using it inside the class will not result in the call of the parent variable.

jaxxzer:
aha, it's clear now.

PR #528, opened 2019-07-02
Final integration with Ping360

patrickelectric:
Needs: https://github.com/bluerobotics/ping-cpp/pull/6

### DIFF
--- a/src/devicemanager/devicemanager.cpp
+++ b/src/devicemanager/devicemanager.cpp
@@ -23,8 +23,8 @@ DeviceManager::DeviceManager() :
         qCDebug(DEVICEMANAGER) << "Available devices:" << links;
         updateAvailableConnections(links);
     });
-    append({AbstractLinkNamespace::Ping1DSimulation}, "Ping1D");
-    append({AbstractLinkNamespace::Ping360Simulation}, "Ping360");
+    append({AbstractLinkNamespace::Ping1DSimulation, {}, "Ping1D Simulation", PingDeviceType::PING1D}, "Ping1D");
+    append({AbstractLinkNamespace::Ping360Simulation, {}, "Ping360 Simulation", PingDeviceType::PING360}, "Ping360");
 }
 
 void DeviceManager::append(const LinkConfiguration& linkConf, const QString& deviceName)
@@ -90,12 +90,12 @@ void DeviceManager::connectLink(LinkConfiguration* linkConf)
         return;
     }
 
+    _sensors[Name][objIndex] = PingHelper::nameFromDeviceType(linkConf->deviceType());
     qCDebug(DEVICEMANAGER) << "Connecting with sensor:" << _sensors[Name][objIndex].toString() << *linkConf;
 
     // We could use a single Ping instance, but since we are going to support multiple devices
     // this pointer will hold everything for us
-    // TODO: Use enum and not string compare
-    if(_sensors[Name][objIndex].toString() == "Ping1D") {
+    if(linkConf->deviceType() == PingDeviceType::PING1D) {
         _primarySensor.reset(new Ping());
     } else {
         _primarySensor.reset(new Ping360());
@@ -131,9 +131,8 @@ void DeviceManager::updateAvailableConnections(const QVector<LinkConfiguration>&
         emit dataChanged(indexRow, indexRow, _roles);
     }
 
-    // Right now we only support Ping1D, we should update protocol detector to detect device type
     for(const auto& link : availableLinkConfigurations) {
-        append(link, "Ping1D");
+        append(link, PingHelper::nameFromDeviceType(link.deviceType()));
     }
 }
 
--- a/src/link/linkconfiguration.cpp
+++ b/src/link/linkconfiguration.cpp
@@ -119,13 +119,19 @@ bool operator==(const LinkConfiguration& first, const LinkConfiguration& second)
     return (firstLinkconf.name == secondLinkconf.name)
            && (firstLinkconf.type == secondLinkconf.type)
            && (firstLinkconf.args == secondLinkconf.args)
+           && (firstLinkconf.deviceType == secondLinkconf.deviceType)
            ;
 }
 
 QDebug operator<<(QDebug d, const LinkConfiguration& other)
 {
-    QString text(QStringLiteral("LinkConfiguration{Name: %1, LinkType: %2, Arguments: (%3)}"));
-    d << text.arg(other.name(), QString::number(other.type()), other.args()->join(":"));;
+    QString text(QStringLiteral("LinkConfiguration{Name: %1, Sensor: %2, LinkType: %3, Arguments: (%4)}"));
+    d << text.arg(
+          other.name(),
+          PingHelper::nameFromDeviceType(other.deviceType()),
+          other.typeToString(),
+          other.args()->join(":")
+      );
     return d;
 }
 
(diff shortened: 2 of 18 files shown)


src/sensor/ping360.h
```diff
@@ -274,15 +364,19 @@ class Ping360 : public PingSensor
...
     uint32_t _speed_of_sound = 1500;
     uint16_t _transmit_frequency = 500;
 ///@}
 
-    uint16_t _num_points = 2000;
+    int _angle_offset = 0;
+    int _angular_speed = 1;
+    uint _central_angle = 1;
+    uint16_t _num_points = 1000;
+    bool _reverse_direction = false;
```
jaxxzer: My preference is to call this variable _direction, and handle it as the inverse, so you don't have to apply that double-negative logic in your head when you read the code.

If we want to be really c++ with it and clear we could have an
```
enum ScanDirection {
  NORMAL,
  REVERSE,
}
```

jaxxzer: doesn't hold pr.

src/sensor/ping360.h
```diff
@@ -236,6 +236,92 @@ class Ping360 : public PingSensor
     }
     Q_PROPERTY(int speed_of_sound READ speed_of_sound WRITE set_speed_of_sound NOTIFY speedOfSoundChanged)
 
+    /**
+     * @brief Returns the angle offset from sample position
+     *
+     * @return int
+     */
+    int angle_offset() { return _angle_offset; }
```
jaxxzer: Consider if this should be a property of the visualizer, so we can do a transform of all the existing data inside the visualizer buffer.

src/waterfall/polarplot.cpp
```diff
@@ -74,6 +74,7 @@ void PolarPlot::draw(const QVector<double>& points, float angle, float initPoint
     static float angleResolution = gradianToDegree/2;
     float actualAngle = angle*gradianToRadian;
 
+    //TODO: Need a better way to deal with dynamic steps
```
jaxxzer: An idea: instead of adding angleGrad parameter to this function, the caller could just loop over the sweep angle, and call `draw(data, angle++)`. This will sacrifice performance in calls to draw, but we have to meet that performance requirement anyway when the step resolution == 1, so I don't think that sacrifice matters.

src/sensor/protocoldetector.cpp
```diff
@@ -190,18 +190,26 @@ bool ProtocolDetector::checkBuffer(const QByteArray& buffer)
...
+                                                    << "\ndevice_type:" << device_information.device_type()
+                                                    << "\ndevice_revision:" << device_information.device_revision()
+                                                    << "\nfirmware_version_major:" << device_information.firmware_version_major()
+                                                    << "\nfirmware_version_minor:" << device_information.firmware_version_minor()
+                                                    << "\nfirmware_version_patch:" << device_information.firmware_version_patch();
+
+            //TODO: Ping1D with firmware 3.26 or older
+            // We should remove this code in future releases but allowing undetected devices to be flashed
+            PingDeviceType type = static_cast<PingDeviceType>(device_information.device_type());
+            if(type == PingDeviceType::PING360) {
```
jaxxzer: IMO, this is an example where implicit conversions are desirable (less code), but we can't with `enum class`.

src/sensor/protocoldetector.cpp
```diff
@@ -190,18 +190,26 @@ bool ProtocolDetector::checkBuffer(const QByteArray& buffer)
...
-                                                << "\nfirmware_version_major:" << device_information.firmware_version_major()
-                                                << "\nfirmware_version_minor:" << device_information.firmware_version_minor()
-                                                << "\nfirmware_version_patch:" << device_information.firmware_version_patch();
-    });
-
+    qCDebug(PING_PROTOCOL_PROTOCOLDETECTOR) << buffer;
     for (const auto& byte : buffer) {
         if(parser.parseByte(byte) == Parser::NEW_MESSAGE) {
+            // Print information from detected devices
+            common_device_information device_information(parser.rxMessage());
```
jaxxzer: To be safe, we should verify the message id before the copy. It might be a Nack or something unexpected.

src/devicemanager/devicemanager.cpp
```diff
@@ -133,7 +134,7 @@ void DeviceManager::updateAvailableConnections(const QVector<LinkConfiguration>&
 
     // Right now we only support Ping1D, we should update protocol detector to detect device type
```
jaxxzer: stale comment?

src/devicemanager/devicemanager.cpp
```diff
@@ -90,12 +90,13 @@ void DeviceManager::connectLink(LinkConfiguration* linkConf)
         return;
     }
 
+    _sensors[Name][objIndex] = PingHelper::nameFromDeviceType(linkConf->deviceType());
     qCDebug(DEVICEMANAGER) << "Connecting with sensor:" << _sensors[Name][objIndex].toString() << *linkConf;
 
     // We could use a single Ping instance, but since we are going to support multiple devices
     // this pointer will hold everything for us
     // TODO: Use enum and not string compare
```
jaxxzer: stale comment?

src/link/linkconfiguration.cpp
```diff
@@ -134,16 +140,24 @@ QDataStream& operator<<(QDataStream &out, const LinkConfiguration linkConfigurat
     out << linkConfiguration.configurationStructPtr()->name;
     out << QVariant(linkConfiguration.configurationStructPtr()->args);
     out << QVariant(linkConfiguration.configurationStructPtr()->type);
+    //TODO: We need to register PingDeviceType
+    out << QVariant(static_cast<int>(linkConfiguration.configurationStructPtr()->deviceType));
     return out;
 }
 
 QDataStream& operator>>(QDataStream &in, LinkConfiguration &linkConfiguration)
```
jaxxzer: Is this operator used anywhere?

src/link/linkconfiguration.h
```diff
@@ -24,6 +25,11 @@ class LinkConfiguration : public QObject
         QStringList args;
         QString name;
         LinkType type = LinkType::None;
+        //TODO: We should somehow make this class or this structure more abstract
+        // and remove any protocol or device specific information
+        // right now this is necessary, since link configuration is the our default
+        // link information structure
+        PingDeviceType deviceType = PingDeviceType::UNKNOWN;
```
jaxxzer: Can you remind me briefly why this is necessary @patrickelectric?

src/sensor/protocoldetector.cpp
patrickelectric: This will be fixed after moving ping-message class to enum class ;)

src/sensor/protocoldetector.cpp
patrickelectric: Ping1D will reply with a NACK message, that's why I'm not checking. Will see that I can do related to it, otherwise a TODO will be add;

src/link/linkconfiguration.cpp
patrickelectric: settingsmanager.h:107

src/link/linkconfiguration.h
patrickelectric: This is necessary to share device identification from protocoldetector to devicemanager.

src/sensor/ping360.h
patrickelectric: Added TODO.

src/waterfall/polarplot.cpp
patrickelectric: TODO updated.

src/sensor/ping360.h
patrickelectric: I think we should do that in the entire pingviewer. Will add an issue for that.

src/sensor/protocoldetector.cpp
jaxxzer: The suggestion is this: 
```
if (parser.rxMessage()->message_id() == CommonId::DEVICE_INFORMATION) {
    common_device_information device_information(parser.rxMessage());
    ...
}
```

You should not cast a pingmessage to a child type before checking the message id matches the child type.

src/sensor/protocoldetector.cpp
jaxxzer: I think you missed my point: 

> IMO, this is an example where implicit conversions are desirable (less code), but **we can't with enum class.**

I don't like `enum class` right here, it is not the right place for `enum class`.

IMO:
`enum class` is good to use sometimes, but not here. In this particular situation, implicit conversions are **desireable** and better. `enum class` does not allow implicit conversions. This is part of the reason I have been recommending against using it for the ping-message api.

patrickelectric: Implicit conversion is only necessary because ping-message does not provide the enums. The problem is in ping-message and not ping-viewer. Implicit conversions are necessary in pingviewer because ping-message does not. It's not safe to do this kind of conversions, is not save for the programmer to do this kind of conversions, we should do it in our low level API not here. Yes, right now we need to do implicit conversions, but this is an API problem in ping-message and not pingviewer. If you dislike implicit conversions, this should be fixed in ping-message providing the correct output as enum classes.

jaxxzer: > If you dislike implicit conversions

I **do like** the **implicit** conversions. That is my point. The implicit conversion is safe in this case because message id is an integer **by definition**.

PR #533, opened 2019-07-05
Ping360: Allow dynamic size array with preallocated space 

patrickelectric:


src/sensor/ping360.cpp
```diff
@@ -25,7 +25,7 @@ Q_LOGGING_CATEGORY(PING_PROTOCOL_PING360, "ping.protocol.ping360")
 
 Ping360::Ping360()
     :PingSensor()
-    ,_data(_num_points, 0)
+    ,_data(_sensorTimeout, 0)
```
jaxxzer: Did you mean to use `_sensorTimeout` to initialize here? That does not look correct.

src/sensor/ping360.cpp
```diff
@@ -93,8 +93,8 @@ void Ping360::handleMessage(const ping_message& msg)
         _transmit_duration = deviceData.transmit_duration();
         _sample_period = deviceData.sample_period();
         _transmit_frequency = deviceData.transmit_frequency();
-        deviceData.number_of_samples();
 
+        _data.resize(deviceData.number_of_samples());
```
jaxxzer: Why do we need to dynamically resize the buffer? It appears to be an unnecessary operation that will hurt performance. I think we should allocate the maximum expected size one time in the constructor.

patrickelectric: We prealocate 2k memory, resize will only change the `size()` return of the class.

src/sensor/ping360.cpp
patrickelectric: yep, should be `_maxNumberOfPoints`.

src/sensor/ping360.cpp
jaxxzer: [QByteArray::resize()](https://doc.qt.io/qt-5/qbytearray.html#resize) looks like it might do a reallocation every time. @tcanabrava do you know if this is an expensive operation?

tcanabrava:
It is but it’s way less expensive than to append each item in a for loop.


Em qua, 17 de jul de 2019 às 22:45, Jacob Walser <notifications@github.com>
escreveu:

src/sensor/ping360.cpp
jaxxzer: I made a mistake `_data` here is QVector, not QByteArray: https://doc.qt.io/qt-5/qvector.html#resize

PR #534, opened 2019-07-05
polarplot: Correct mouse distance and angle to be degrees and meters

patrickelectric:

### DIFF
--- a/src/waterfall/polarplot.cpp
+++ b/src/waterfall/polarplot.cpp
@@ -113,8 +113,8 @@ void PolarPlot::draw(const QVector<double>& points, float angle, float initPoint
 
 void PolarPlot::updateMouseColumnData()
 {
-    static const QPoint center(width()/2, height()/2);
     static const float rad2grad = 200/M_PI;
+    const QPointF center(width()/2, height()/2);
 
     const QPoint delta = _mousePos - center;
     // Check if mouse is inside circle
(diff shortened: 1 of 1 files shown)


src/waterfall/polarplot.cpp
```diff
@@ -127,9 +129,10 @@ void PolarPlot::updateMouseColumnData()
     // Calculate grad from mouse position
     const QPoint deltaScaled(delta.x()*_image.width()/width(), delta.y()*_image.height()/height());
     int grad = static_cast<int>(atan2f(-deltaScaled.x(), deltaScaled.y())*rad2grad + 200) % 400;
+    _mouseSampleAngle = grad*grad2deg;
+    // Claculate mouse distance in meters
```
Williangalvani: Typo

src/waterfall/polarplot.cpp
```diff
@@ -127,9 +129,10 @@ void PolarPlot::updateMouseColumnData()
...
+    _mouseSampleDistance = 2*std::hypotf(delta.x(), delta.y())/std::max(width(), height())*_maxDistance*1e-3;
```
Williangalvani: Can you add a comment explaining this? What I got is that you are using the factor between the radius of the current mouse position and the full circle multiplied by the resolution in meters. But then why the 2* and why not `min()` instead of `max()`?

src/waterfall/polarplot.cpp
patrickelectric: Fixed.

src/waterfall/polarplot.cpp
patrickelectric: The `2` comes from the division of the width|height, so it goes up as a multiplication, will add the equations to be clear.
The `std::` call was removed in the improved version.

PR #535, opened 2019-07-08
Update Visualizer function and clear plot in scale changes 

patrickelectric:
deepin-screen-recorder_pingviewer_20190708141526

### DIFF
--- a/qml/Ping360Visualizer.qml
+++ b/qml/Ping360Visualizer.qml
@@ -22,6 +22,13 @@ Item {
         property var ping: DeviceManager.primarySensor
         target: ping
 
+        /** Ping360 does not handle auto range/scale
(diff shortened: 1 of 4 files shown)


qml/Ping360Visualizer.qml
```diff
@@ -22,6 +22,13 @@ Item {
         property var ping: DeviceManager.primarySensor
         target: ping
 
+        /** Ping360 does not handle auto range/scale
+         *  Any change in scale is a result of user input
+         */
+        onLength_mmChanged: {
```
jaxxzer: I think this is better than the current behavior. In the future, it will be even better if we do not clear the the existing data that is remains inside the plot limits.

PR #537, opened 2019-07-10
protocol-detector: add break condition when changing baudrates, react…

jaxxzer:
…ivate 9600 baud scanning

### DIFF
--- a/src/sensor/protocoldetector.cpp
+++ b/src/sensor/protocoldetector.cpp
@@ -70,6 +70,7 @@ void ProtocolDetector::doScan()
 bool ProtocolDetector::checkLink(LinkConfiguration& linkConf)
 {
     _detected = false;
+    _parser.clearBuffer();
     if(linkConf.type() == LinkType::Udp) {
         checkUdp(linkConf);
     } else if(linkConf.type() == LinkType::Serial) {
@@ -216,15 +217,11 @@ bool ProtocolDetector::checkUdp(LinkConfiguration& linkConf)
 
 bool ProtocolDetector::checkBuffer(const QByteArray& buffer, LinkConfiguration& linkConf)
 {
-    // Parser need to be here, having a single parser to for everything will result in fake detections,
-    // since the buffer need to be clear for each try
-    PingParserExt parser;
-
     qCDebug(PING_PROTOCOL_PROTOCOLDETECTOR) << buffer;
     for (const auto& byte : buffer) {
-        if(parser.parseByte(byte) == Parser::NEW_MESSAGE) {
+        if(_parser.parseByte(byte) == Parser::NEW_MESSAGE) {
             // Print information from detected devices
-            common_device_information device_information(parser.rxMessage());
+            common_device_information device_information(_parser.rxMessage());
             qCDebug(PING_PROTOCOL_PROTOCOLDETECTOR) << "Detect new device:"
                                                     << "\ndevice_type:" << device_information.device_type()
                                                     << "\ndevice_revision:" << device_information.device_revision()
--- a/src/sensor/protocoldetector.cpp
+++ b/src/sensor/protocoldetector.cpp
@@ -101,7 +101,7 @@ QVector<LinkConfiguration> ProtocolDetector::updateLinkConfigurations(QVector<Li
         // Add valid port and baudrate
         // Ping360 can't handle 9600 requests with 115200 request in a sort time priod
         // TODO: Fix Ping360 is not possible, we should drop 9600 checks if 115200 returns fine
-        for(auto baud : {115200, /*9600*/}) {
+        for(auto baud : {115200, 9600}) {
             auto config = {portInfo.portName(), QString::number(baud)};
             tempConfigs.append({LinkType::Serial, config, QString("Detector serial link")});
         }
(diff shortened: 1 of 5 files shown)


src/sensor/protocoldetector.cpp
```diff
@@ -130,6 +130,12 @@ bool ProtocolDetector::checkSerial(LinkConfiguration& linkConf)
     }
     qCDebug(PING_PROTOCOL_PROTOCOLDETECTOR) << "Port is open";
     port.setBaudRate(baudrate);
+    port.sendBreak();
```
patrickelectric: `sendBreak` is deprecated, please do not use it.

jaxxzer: I don't see anything about deprecation in 5.13: https://doc.qt.io/qt-5/qserialport.html#sendBreak

tcanabrava: it was deprecated, then undeprecated. but patrick is right - this is blocking, and thus should be avoided. https://forum.qt.io/topic/90664/qserialport-sendbreak-is-deprecated-but-no-alternative-is-proposed/5

src/sensor/protocoldetector.cpp
```diff
@@ -214,15 +215,11 @@ bool ProtocolDetector::checkUdp(LinkConfiguration& linkConf)
...
 bool ProtocolDetector::checkBuffer(const QByteArray& buffer, LinkConfiguration& linkConf)
 {
-    // Parser need to be here, having a single parser to for everything will result in fake detections,
-    // since the buffer need to be clear for each try
-    PingParserExt parser;
-
     qCDebug(PING_PROTOCOL_PROTOCOLDETECTOR) << buffer;
     for (const auto& byte : buffer) {
-        if(parser.parseByte(byte) == Parser::NEW_MESSAGE) {
+        if(_parser.parseByte(byte) == Parser::NEW_MESSAGE) {
```
patrickelectric: This will use the same parser in different sensors and connections.

src/sensor/protocoldetector.cpp
jaxxzer: This class runs in it's own thread.

- I think that means it should be ok if it is blocking?
- If we use setbreakenabled and QThread::msleep that is still blocking, correct?
- What is the non-blocking alternative?

patrickelectric: @tcanabrava 	Where did you get that is not deprecated anymore ? Here I have warning messages for both 5.12 and 5.13

jaxxzer: Guys, any suggestion here to get it merged? This fixes bugs preventing proper operation of the device.

patrickelectric: This is results in deprecated warnings here, so I'm a bit apprehensive about it.

jaxxzer: I'll try setbreakenabled with msleep.

jaxxzer: > This results in deprecated warnings here, so I'm a bit apprehensive about it.

can you share the build log and point the line #?

patrickelectric:
```
../src/link/seriallink.cpp: In member function ‘virtual bool SerialLink::setConfiguration(const LinkConfiguration&)’:
../src/link/seriallink.cpp:50:21: warning: ‘bool QSerialPort::sendBreak(int)’ is deprecated [-Wdeprecated-declarations]
   50 |     _port.sendBreak();
      |                     ^
In file included from /usr/include/qt/QtSerialPort/QSerialPort:1,
                 from ../src/link/seriallink.h:3,
                 from ../src/link/seriallink.cpp:7:
/usr/include/qt/QtSerialPort/qserialport.h:269:24: note: declared here
  269 |     QT_DEPRECATED bool sendBreak(int duration = 0);

```

jaxxzer: using setbreakenabled now

patrickelectric:
Is this supposed to work with 2M baud rate ? I got it working with all the important ones but not with 2M.
image

jaxxzer:
> Is this supposed to work with 2M baud rate ? I got it working with all the important ones but not with 2M.

I have encountered the same problem as you. It can work with 2M in the future, that will come in a next PR. In this version, only 115200 is supported, the same as master.

jaxxzer:
Review has been addressed.

src/sensor/protocoldetector.cpp
```diff
@@ -130,6 +131,14 @@ bool ProtocolDetector::checkSerial(LinkConfiguration& linkConf)
     }
     qCDebug(PING_PROTOCOL_PROTOCOLDETECTOR) << "Port is open";
     port.setBaudRate(baudrate);
+    port.setBreakEnabled(true);
+    QThread::usleep(500);
+    port.setBreakEnabled(false);
+    QThread::msleep(11);
```
Williangalvani: Where do these numbers come from? are they the same that Gordon uses?

jaxxzer: They are the same that gordon uses. these numbers will be documented for the product. This routine should also be compatible with any device.

PR #539, opened 2019-07-11
Add sector view

patrickelectric:
video in slack

### DIFF
--- a/qml/Ping360Visualizer.qml
+++ b/qml/Ping360Visualizer.qml
@@ -17,15 +17,19 @@ Item {
     id: root
     property alias displaySettings: displaySettings
     anchors.fill: parent
+    property var ping: DeviceManager.primarySensor
 
     Connections {
-        property var ping: DeviceManager.primarySensor
         target: ping
 
         /** Ping360 does not handle auto range/scale
          *  Any change in scale is a result of user input
          */
         onLength_mmChanged: {
+            clear()
+        }
+
+        onSectorSizeChanged : {
             clear()
         }
 
@@ -51,14 +55,18 @@ Item {
         anchors.fill: parent
 
         Item {
+            id: waterfallParent
             Layout.fillHeight: true
             Layout.fillWidth: true
 
             PolarPlot {
                 id: waterfall
-                height: Math.min(parent.height, parent.width)
+                height: Math.min(ping.sectorSize > 200 ? parent.height : parent.height*2, parent.width*scale)
                 width: height
-                anchors.centerIn: parent
+                anchors.horizontalCenter: parent.horizontalCenter
+                anchors.verticalCenter: ping.sectorSize > 200 ? parent.verticalCenter : parent.bottom
+
+                property var scale: ping.sectorSize >= 200 ? 1 : 0.8/Math.sin(ping.sectorSize*Math.PI/400)
                 property bool verticalFlip: false
                 property bool horizontalFlip: false
 
@@ -159,6 +167,7 @@ Item {
             PolarGrid {
                 id: polarGrid
                 anchors.fill: waterfall
+                angle: ping.sectorSize*360/400
                 maxDistance: waterfall.maxDistance
             }
         }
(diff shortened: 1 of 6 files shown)


src/sensor/ping360.h
```diff
@@ -311,6 +314,28 @@ class Ping360 : public PingSensor
     }
     Q_PROPERTY(int number_of_points READ number_of_points WRITE set_number_of_points NOTIFY numberOfPointsChanged)
 
+    /**
+     * @brief Return sector size in gradians
+     *
+     * @return int
+     */
+    int sectorSize() { return _sectorSize; }
```
jaxxzer: Suggestion for the future: rather than using properties for 'central angle' and 'sector size', I think it _might_ be easier to handle all of this math with ping properties of 'scan start angle' and 'scan end angle'.

jaxxzer:
It will be nice if we can get the graph marks properly aligned with the waterfall data.
Screenshot_20190712_124920

jaxxzer:
There appears to be some problem with selecting different angular resolutions
image

jaxxzer:
Please rebase and test with the angle offset/flip.

jaxxzer:
At each edge of a sector, when the direction changes, the current sector indicator pauses for two sample periods instead of one. It appears that the profile at the edge is skipped, or that it is plotted twice.

kklemens:
We are also going to need a 120 degree sector for the video. From the slack video, it looks like the options we had were 360, 180, 90, 60, 45, 30.

patrickelectric:
Hi Kevin, thank you for your input. But you can edit the sector width, we
have only some basic ones, but you should edit the field without problems.

patrickelectric:
> It will be nice if we can get the graph marks properly aligned with the waterfall data.

`fixup!` fix thats.

> There appears to be some problem with selecting different angular resolutions

This will happen with big steps between samples, will try to fix that.

qml/PolarGrid.qml
```diff
@@ -34,9 +36,9 @@ Item {
         delegate: Text {
             font.pixelSize: 15
             // Calculate the angle and radius for each item
-            property var itemRadian: (index%Math.floor(360/repeater.angle))*repeater.angle*Math.PI/180
-            property var itemRadius: repeater.radius*(1 + Math.floor(index/Math.floor(360/repeater.angle)))/repeater.numberOfRings
-            text: repeater.formatDistance(root.maxDistance*(1 + Math.floor(index/Math.floor(360/repeater.angle)))/repeater.numberOfRings)
+            property var itemRadian: ((root.angle/2)%90)*Math.PI/180 - ((root.angle%180) ? Math.PI/2 : 0)
```
jaxxzer: Some of the math in this file is pretty complicated to understand what/why you are doing. Some comments in the future will help new eyes :eyes:.

qml/PolarGrid.qml
```diff
@@ -62,56 +67,59 @@ Item {
...
             }
         }
     }
 
     Canvas {
+        id: canvas
         anchors.fill: parent
         opacity: 0.8
-        property var angle: 45
+        property var angle: root.angle/45 > 3 ? 45 : root.angle/4
```
jaxxzer: :eyes:

qml/Ping360Visualizer.qml
```diff
@@ -204,6 +213,21 @@ Item {
...
             columnSpacing: 5
 
+            // Sensor connections should be done inside component
+            // Components are local '.qml' descriptions, so it's not possible get outside connections
+            Connections {
+                target: ping
+                // We do not support vertical flips in sector view
+                onSectorSizeChanged: {
+                    if(ping.sectorSize != 400) {
+                        verticalFlipChB.checked = false
```
jaxxzer: :ok_hand:

jaxxzer:
> fixup! fix thats.

Ok, it did help with the 30 degree setting, but the 60 and 45 are still off.

That's ok for now, but we should try to rework things so we have better control and synchronization between the polar plot components and they align down to the pixel.

PR #540, opened 2019-07-11
ping360 scan speed optimizations

jaxxzer:
This is based on #537.

This is the situation that we are dealing with regarding communication between the ping360 device hardware and the ping-viewer application:

image

This time period of 18 milliseconds is what I will term the `viewer-side latency`. The answer to the question 'is this viewer-side latency necessary?' is **no**! Ideally, it will be zero.

The sonar response with default settings is a constant 8ms. The maximum theoretical message frequency @2Mbaud in this case is 1/0.008 = 125Hz (viewer-side latency = 0). The maximum theoretical full scan speed is 400 / 125 = 3.2 seconds

In an attempt to reduce the signal->slot chain between ping-viewer rx -> ping-viewer tx, I did two primary things:

- removed use of newMessage signal from parser, and parse the buffer/handle the message synchronously
- added writeSync to write immediately when abstractlink::write is called

I also moved some calls around inside of `handleMessage` in order to request the next data message from the ping360 device as soon as possible.

**master communicating at default 2m range and 2Mbaud scans at 30 Hz**
**this patch communicating at default 2m range and 2Mbaud scans at 60 Hz**

There remains a ton of jitter from the async (maybe readyRead())?, sometimes the viewer-side latency is 20ms.

As a point of reference for what is technically possible, the minimal example [test-device-ping360.cpp](https://github.com/jaxxzer/ping-cpp/blob/test-device/test/test-device-ping360.cpp) has a full scan time of 3.6 seconds (110Hz) and the viewer-side latency is always less than 1ms.

### DIFF
--- a/src/link/abstractlink.h
+++ b/src/link/abstractlink.h
@@ -212,9 +212,23 @@ public:
     void write(const char* data, int size)
     {
         if(size > 0) {
-            emit sendData(QByteArray(data, size));
-        };
+            if (_type == LinkType::Serial){
+                writeSync(data, size);
+            } else {
+                emit sendData(QByteArray(data, size));
+            }
+        }
     }
+
+    virtual void writeSync(const char* data, int size) { Q_UNUSED(data) Q_UNUSED(size) }
+
+    /**
+     * @brief Copy operation
+     *
+     * @param other
+     * @return const AbstractLink&
+     */
+    const AbstractLink& operator=(const AbstractLink& other);
 
     Q_PROPERTY(qint64 byteSize READ byteSize NOTIFY byteSizeChanged)
     Q_PROPERTY(LinkConfiguration* configuration READ configuration NOTIFY configurationChanged)
--- a/src/link/abstractlink.h.orig
+++ b/src/link/abstractlink.h.orig
@@ -212,7 +212,7 @@ public:
     void write(const char* data, int size)
     {
         if(size > 0) {
-            if (_type == LinkType::Serial) {
+            if (_type == LinkType::Serial){
                 writeSync(data, size);
             } else {
                 emit sendData(QByteArray(data, size));
--- a/src/sensor/ping360.h
+++ b/src/sensor/ping360.h
@@ -543,7 +543,7 @@ private:
     int _angular_speed = 1;
     bool _autoTransmitDuration = true;
     uint _central_angle = 1;
-    bool _configuring = true;
+    bool _configuring = false;
     // Number of messages used to check the best baud rate
     int _preConfigurationTotalNumberOfMessages = 20;
     bool _reverse_direction = false;
--- a/src/sensor/pingsensor.cpp
+++ b/src/sensor/pingsensor.cpp
@@ -14,6 +14,21 @@ PingSensor::PingSensor()
     connect(dynamic_cast<PingParserExt*>(_parser), &PingParserExt::parseError, this, &PingSensor::parserErrorsUpdate);
 }
 
+void PingSensor::connectLink(const LinkConfiguration conConf, const LinkConfiguration& logConf)
+{
+    Sensor::connectLink(conConf, logConf);
+    if (_parser) {
+        connect(link(), &AbstractLink::newData, this, [this](QByteArray data) {
+            for (char b : data) {
+                Parser::ParserState result = _parser->parseByte(b);
+                if (result == Parser::NEW_MESSAGE) {
+                    handleMessage(_parser->rxMessage());
+                }
+            }
+        });
+        emit this->connectionUpdate();
+    }
+}
 
 void PingSensor::request(int id) const
 {
--- a/src/link/seriallink.h
+++ b/src/link/seriallink.h
@@ -92,6 +92,11 @@ public:
      */
     void forceSensorAutomaticBaudRateDetection();
 
+    void writeSync(const char* data, int size) final override {
+        _port.write(data, size);
+        _port.flush();
+    }
+
 private:
     QSerialPort _port;
 };
--- a/src/link/seriallink.h.orig
+++ b/src/link/seriallink.h.orig
@@ -92,8 +92,7 @@ public:
      */
     void forceSensorAutomaticBaudRateDetection();
 
-    void writeSync(const char* data, int size) final override
-    {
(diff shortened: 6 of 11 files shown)


jaxxzer:
fix #650

patrickelectric:
@jaxxzer If you can, please break this in multiple PRs to make it easier to master.

jaxxzer:
fix #657

jaxxzer:
> @jaxxzer If you can, please break this in multiple PRs to make it easier to master.

I cannot. This patch only works when applied as a whole.

I have updated the commit messages and comments to try to make it easy. Please let me know if you have any trouble understanding.

The hotfix is applied to get around #656

jaxxzer:
`master` has scan speed of 30Hz

This branch has scan speed of 60Hz.

src/link/abstractlink.h
```diff
@@ -212,10 +212,24 @@ class AbstractLink : public QObject
     void write(const char* data, int size)
     {
         if(size > 0) {
-            emit sendData(QByteArray(data, size));
-        };
+            if (_type == LinkType::Serial){
```
jaxxzer: I put this case here because I have not tested other link types. I think that maybe we should favor the `writeSync` logic over the `emit sendData()` logic for everyone.

jaxxzer:
@patrickelectric Please check also the updated [pr description](https://github.com/bluerobotics/ping-viewer/pull/540#issue-296843485).

src/link/abstractlink.h.orig
```diff
@@ -0,0 +1,269 @@
+#pragma once
```
patrickelectric: Please, remove all .orig files.

src/link/abstractlink.h
```diff
@@ -212,10 +212,24 @@ class AbstractLink : public QObject
     void write(const char* data, int size)
     {
         if(size > 0) {
-            emit sendData(QByteArray(data, size));
-        };
+            if (_type == LinkType::Serial) {
+                writeSync(data, size);
```
patrickelectric: why don't you make this the default behavior of the serial port ?
If this is link specific should be in the link and not in abstract link.
If you are trying to avoid the eventloop signal-slot handle, you just need to make the connect attribute as direct and not queue.

src/sensor/pingsensor.cpp
```diff
@@ -14,6 +14,21 @@ PingSensor::PingSensor()
     connect(dynamic_cast<PingParserExt*>(_parser), &PingParserExt::parseError, this, &PingSensor::parserErrorsUpdate);
 }
 
+void PingSensor::connectLink(const LinkConfiguration conConf, const LinkConfiguration& logConf)
+{
+    Sensor::connectLink(conConf, logConf);
+    if (_parser) {
```
patrickelectric: Add a !parser check with a warning/critical message, this avoid the extra indentation and helps to debug. Right now it fails in silence.

src/sensor/pingsensor.cpp
```diff
@@ -14,6 +14,21 @@ PingSensor::PingSensor()
...
+        connect(link(), &AbstractLink::newData, this, [this](QByteArray data) {
```
patrickelectric: why are you not using the newMessage signal and doing the parser job here ? This is not explicit described in the commit body or in comments. How this helps ?
Also, should this be fixed in the `parseBuffer` function ?

patrickelectric: You are doing a deep copy of the QByteArray, use reference.

src/sensor/pingsensor.cpp
```diff
@@ -14,6 +14,21 @@ PingSensor::PingSensor()
...
+            for (char b : data) {
+                Parser::ParserState result = _parser->parseByte(b);
+                if (result == Parser::NEW_MESSAGE) {
+                    handleMessage(_parser->rxMessage());
+                }
+            }
+        });
+        emit this->connectionUpdate();
```
patrickelectric: `this` is not necessary.

patrickelectric: Also, how this update the connection ? This is done by the base sensor class, and should be done only when the state of the connection changes, and you are not doing it here.

src/sensor/ping360.h
```diff
@@ -543,7 +543,7 @@ class Ping360 : public PingSensor
     int _angular_speed = 1;
     bool _autoTransmitDuration = true;
     uint _central_angle = 1;
-    bool _configuring = true;
+    bool _configuring = false;
```
patrickelectric: It's working here, how is it broken ? There is an issue for that open ? Is this PR breaking the autobaud logic ?

src/link/abstractlink.h
```diff
@@ -212,7 +212,7 @@ class AbstractLink : public QObject
     void write(const char* data, int size)
     {
         if(size > 0) {
-            if (_type == LinkType::Serial){
+            if (_type == LinkType::Serial) {
```
patrickelectric: This commit should be applied in the correct places.

src/link/abstractlink.h
jaxxzer: > why don't you make this the default behavior of the serial port ?
If this is link specific should be in the link and not in abstract link.

I put this case here because I have not tested other link types. I think that maybe we should favor the writeSync logic over the emit sendData() logic for everyone.

> If you are trying to avoid the eventloop signal-slot handle, you just need to make the connect attribute as direct and not queue.

what magic do you speak of?

src/sensor/pingsensor.cpp
jaxxzer: > removed use of newMessage signal from parser, and parse the buffer/handle the message synchronously

I'm sorry it's not clear, I'm trying to handle the message immediately, rather than waiting for the newMessage signal to propagate. (I'm a noob at threading and signal/slots, my understanding is not solid).

jaxxzer: nice :eye:

src/sensor/pingsensor.cpp
jaxxzer: I think I was trying to maintain some order of calls that was already existing. Probably you are right that this does not belong.

PR #542, opened 2019-07-12
Add offset

patrickelectric:


src/sensor/ping360.h
```diff
@@ -359,14 +359,15 @@ class Ping360 : public PingSensor
     QVector<double> _data;
 ///@}
 
-    int _angle_offset = 0;
+    // Ping360 has a 200 offset by default
+    int _angle_offset = 200;
```
jaxxzer: I think this ought to be a property of the visualizer.

jaxxzer:
This is all wrong to me. You should be using the same transform that you are using with horizontal/vertical flip.

jaxxzer:
There should be no change to the scan/angle logic in ping sensor.

PR #545, opened 2019-07-17
Ping360 debug info

jaxxzer:


patrickelectric:
This should be handle by the sensor class, like we do with control and visualizer. The class should provide a debug widget, this will remove the duplicate code around both sensors.

jaxxzer:
> This should be handle by the sensor class, like we do with control and visualizer. The class should provide a debug widget, this will remove the duplicate code around both sensors.

I agree with you, but this (non-functional) change will take significantly more time to prepare. I have added some TODOs according to your recommendation: https://github.com/bluerobotics/ping-viewer/pull/545/files#diff-2306966e04af59118a3a97b499d607d7R138

qml/Ping360Status.qml
```diff
@@ -0,0 +1,52 @@
...
+                    "Number of samples (#): " + ping.number_of_points,
+                    "Ping (#): " + ping.ping_number,
+                    "Transmit duration (μs): " + ping.transmit_duration,
+                    "Gain (setting): " + ping.gain_setting,
+                    "Speed of sound (mm/s): " + ping.speed_of_sound,
+                    "Lost messages (#): " + ping.lost_messages,
+                    "RX Packets (#): " + ping.parsed_msgs,
+                    "RX Errors (#): " + ping.parser_errors,
+                    "Ascii text:\n" + ping.ascii_text,
+                    "Error message:\n" + ping.err_msg,
```
patrickelectric: It's missing angle, frequency and offset.

qml/Ping360Status.qml
```diff
@@ -0,0 +1,52 @@
+import QtQuick 2.7
```
patrickelectric: What is the meaning of `og` ? 
From google:
image

We should avoid this kind of alias in commits, they are not clear.

jaxxzer: done

qml/Ping360Status.qml
jaxxzer: done

PR #547, opened 2019-07-17
Some random code corrections and stability checks  

patrickelectric:


src/waterfall/polarplot.cpp
```diff
@@ -114,7 +114,7 @@ void PolarPlot::draw(const QVector<double>& points, float angle, float initPoint
 
 void PolarPlot::updateMouseColumnData()
 {
-    static const float rad2grad = 200/M_PI;
+    static const float rad2grad = 200.0f/M_PI;
```
tcanabrava: if you are using M_PI (that's a define, and not a proper value) perhaps you wan't to use boost for that:
boost::math::constants::pi<float>();

src/sensor/ping.cpp
```diff
@@ -233,7 +233,7 @@ void Ping::handleMessage(const ping_message& msg)
 
     case Ping1dId::MODE_AUTO: {
         ping1d_mode_auto m(msg);
-        if(_mode_auto != m.mode_auto()) {
+        if(_mode_auto != static_cast<bool>(m.mode_auto())) {
```
tcanabrava: I don't get this one. is _mode_auto a different type than m.mode_auto() ?

pingviewer.pro
```diff
@@ -22,6 +22,15 @@ RESOURCES += \
...
 
+# Warning as error
+*-g++ | *-clang {
+    QMAKE_CXXFLAGS += -Werror
+}
+*msvc {
+    QMAKE_CXXFLAGS += /WX \
+        /wd4305 \ # Remove truncated warnings, msvc does not provide some non double values
+}
+
```
tcanabrava: ++.
but why are you using qmake for pingviewer?

src/sensor/ping360.h
```diff
@@ -32,7 +32,7 @@ class Ping360 : public PingSensor
      * @param connType connection type
      * @param connString arguments for the new connection
      */
-    Q_INVOKABLE void connectLink(AbstractLinkNamespace::LinkType connType, const QStringList& connString);
+    Q_INVOKABLE void connectLink(AbstractLinkNamespace::LinkType connType, const QStringList& connString) final override;
```
tcanabrava: if it's final it needs override?

patrickelectric: no, but we want to be explicit here.

src/waterfall/polarplot.cpp
patrickelectric: yep, but adding a single (and big) dependency just because of M_PI is not a good approach.

src/sensor/ping.cpp
patrickelectric: yes, right now we only support "auto mode" and "manual mode" in ping1d, so this a bool is used, there's future plans to add more modes and probably that's why mode_auto returns a `uint8_t`.

pingviewer.pro
patrickelectric: https://github.com/bluerobotics/ping-viewer/issues/549

src/sensor/ping.cpp
jaxxzer: A note that this field will be reserved in the near future, and this line will be removed.

PR #561, opened 2019-07-18
Fixes

tcanabrava:


src/devicemanager/devicemanager.cpp
```diff
@@ -97,11 +97,10 @@ void DeviceManager::connectLink(LinkConfiguration* linkConf)
 
     // We could use a single Ping instance, but since we are going to support multiple devices
     // this pointer will hold everything for us
-    if(linkConf->deviceType() == PingDeviceType::PING1D) {
-        _primarySensor.reset(new Ping());
-    } else {
-        _primarySensor.reset(new Ping360());
-    }
+     _primarySensor.reset( linkConf->deviceType() == PingDeviceType::PING1D
+        ? static_cast<PingSensor*>(new Ping())
```
jaxxzer: my preference is the prior if-else for readability.

src/devicemanager/devicemanager.cpp
```diff
@@ -120,7 +120,7 @@ void DeviceManager::connectLinkDirectly(AbstractLinkNamespace::LinkType connType
 void DeviceManager::updateAvailableConnections(const QVector<LinkConfiguration>& availableLinkConfigurations)
 {
     // Make all connections unavailable by default
-    for(int i{0}; i < _sensors[Available].size(); i++) {
+    for(int i{0}, end = _sensors[Available].size(); i < end; i++) {
```
jaxxzer: squash with 1bc4777906c3a3eb8ba578183ecaf37a1ce85ed5

tcanabrava:
I believe that’s just a matter of getting used to the ternary, it has a
better readability than a if else, The only thing I don’t like there is the
static cast l, but in the end it’s good that it displays the list type.


Em qui, 18 de jul de 2019 às 15:20, Jacob Walser <notifications@github.com>
escreveu:

src/flash/flasher.cpp
```diff
@@ -64,17 +64,17 @@ void Flasher::setVerify(bool verify)
...
-#else
-    static QString absoluteBinPath = QCoreApplication::applicationDirPath();
+        + QStringLiteral("../..")
 #endif
+        + QStringLiteral("/stm32flash")
 #ifdef Q_OS_WIN
-    return absoluteBinPath + "/stm32flash.exe";
-#else
-    return absoluteBinPath + "/stm32flash";
+        + QStringLiteral(".exe");
```
jaxxzer: extra semicolon.

src/flash/flasher.cpp
```diff
@@ -104,7 +104,7 @@ void Flasher::flash()
     static QString cmd = QStringLiteral("\"%0\" -w \"%1\" %2 -g 0x0 -b %3 %4").arg(
                              stm32flashPath(),
                              firmwareFileInfo.absoluteFilePath(),
-                             _verify ? "-v" : "",
+                             _verify ? QStringLiteral("-v") : QString(),
```
jaxxzer: What is the reasoning here? Is this potentially a functional change?

src/devicemanager/devicemanager.cpp
tcanabrava: actually both versions should die, PingDeviceType is an enum, and thus we need to use an switch.

src/flash/flasher.cpp
tcanabrava: Same behavior. QStringLiteral() will transform the strings from utf8 or ascii to utf16 in compile time, and "" will produce a empty string (just with an \0 byte) that will not be appended anywhere because it's empty, while QString() will produce.. nothing. not even a \0 byte. it will not try to convert ascii to utf-16 and thus will execute faster.

src/devicemanager/devicemanager.cpp
jaxxzer: ok

jaxxzer:
Terenaries are ok, but you placed it inside the function call arguments. My opinion is that that is less readable.

```
call( condition
  ? case1
  : case2
)
```

I wouln't have made the comment if you had done
```
condition
  ? call(case1)
  : call(case2)
```

tcanabrava:
Those are the same thing: a return value will be used in place. While an if
creates an scoped area, and no return value.

Rust fixed that by transforming an it into something that can return values.


Em qui, 18 de jul de 2019 às 15:50, Jacob Walser <notifications@github.com>
escreveu:

src/link/link.h
```diff
@@ -37,7 +37,7 @@ class Link : public QObject
      *
      * @return AbstractLink*
      */
-    AbstractLink* self() { _abstractLink };
+    AbstractLink* self() { return _abstractLink; };
```
jaxxzer: this belongs to ca39232f92220c6174344a4eaf146a5635c692b6

src/link/seriallink.cpp
```diff
@@ -61,17 +61,18 @@ bool SerialLink::finishConnection()
...
+        currentPortList.append(port.portName());
     }
-    if(oldList != list) {
+
+    if(_availableConnections != currentPortList) {
         emit availableConnectionsChanged();
     }
-    return list;
+
+    return _availableConnections;
```
jaxxzer: The `_availableConnections` value is never updated.

src/network/networktool.cpp
```diff
@@ -12,39 +12,39 @@
...
     //*github.com/user/repo* results in user/repo
     const static QRegularExpression regex(R"(github.com\/([^.]*))");
-    QRegularExpressionMatch regexMatch = regex.match(QStringLiteral(GIT_URL));
+    QRegularExpressionMatch regexMatch = regex.match(userRepoUrl);
     if (!regexMatch.hasMatch() || regexMatch.lastCapturedIndex() < 1) {
-        qCWarning(NETWORKTOOL) <<
-                               "Fail to get github user and repository! "
-                               "It will not be possible to check for updates. "
-                               "Using default value:" << _gitUserRepo;
+        qCWarning(NETWORKTOOL) << "Fail to get github user and repository! "
```
jaxxzer: should this one be qstringliteral?

tcanabrava: no, qDebug and friends don't need a QString and work witha  char *

src/link/seriallink.cpp
tcanabrava: ups. :)

jaxxzer:
@tcanabrava please ping me when the review is addressed and the ci is fixed (run astyle).

src/flash/flasher.cpp
```diff
@@ -12,15 +12,14 @@
 PING_LOGGING_CATEGORY(FLASH, "ping.flash")
 
 Flasher::Flasher(QObject* parent)
-    : QObject(parent)
-{
-    _binRelativePath =
+    : QObject(parent),
+    _binRelativePath(QCoreApplication::applicationDirPath()
```
patrickelectric: use comma in the next line to make it better to read and avoid future unnecessary diffs.
```
: blah
, foo
, bee
```

src/network/networktool.cpp
```diff
@@ -12,39 +12,39 @@
...
-QString NetworkTool::_gitUserRepoFirmware = "bluerobotics/ping-firmware";
-
+namespace {
+    QString gitUserRepo = QStringLiteral("bluerobotics/ping-viewer");
+    QString gitUserRepoFirmware = QStringLiteral("bluerobotics/ping-firmware");
+    const QString gitReleases = QStringLiteral("https://api.github.com/repos/%1/releases");
+    const QString repoContents = QStringLiteral("https://api.github.com/repos/%1/contents/%2");
+    const QString projectTag = QStringLiteral(GIT_TAG);
+    const QString userRepoUrl = QStringLiteral(GIT_URL);
+}
```
patrickelectric: Add a space between namespace and class constructor, also the commit does not describes what is happening here and why this namespace is necessary.

src/network/networktool.cpp
```diff
@@ -12,39 +12,39 @@
...
 PING_LOGGING_CATEGORY(NETWORKTOOL, "ping.networktool")
 
-
-QString NetworkTool::_gitUserRepo = "bluerobotics/ping-viewer";
-QString NetworkTool::_gitUserRepoFirmware = "bluerobotics/ping-firmware";
-
+namespace {
+    QString gitUserRepo = QStringLiteral("bluerobotics/ping-viewer");
+    QString gitUserRepoFirmware = QStringLiteral("bluerobotics/ping-firmware");
+    const QString gitReleases = QStringLiteral("https://api.github.com/repos/%1/releases");
```
patrickelectric: maybe: `githubReleasesApiPath`to be more clear, just gitReleases does not imply that the variable is a static path for an API.

src/network/networktool.cpp
```diff
@@ -12,39 +12,39 @@
 
 PING_LOGGING_CATEGORY(NETWORKTOOL, "ping.networktool")
 
-
-QString NetworkTool::_gitUserRepo = "bluerobotics/ping-viewer";
-QString NetworkTool::_gitUserRepoFirmware = "bluerobotics/ping-firmware";
-
+namespace {
+    QString gitUserRepo = QStringLiteral("bluerobotics/ping-viewer");
+    QString gitUserRepoFirmware = QStringLiteral("bluerobotics/ping-firmware");
```
patrickelectric: const ?

src/link/filelink.cpp
```diff
@@ -104,9 +104,7 @@ bool FileLink::isOpen()
 bool FileLink::finishConnection()
 {
     // Only close files that are open
-    if(_file.isOpen()) {
-        _file.close();
-    }
+    _file.close();
```
patrickelectric: IIRC this will result in a warning message saying that it's closing something that is not open.

src/link/link.cpp
```diff
@@ -37,7 +37,7 @@ Link::Link(LinkType linkType, QString name, QObject* parent)
         _abstractLink.reset(new Ping360SimulationLink());
         break;
     default :
-        qCDebug(PING_PROTOCOL_LINK) << "Link not available!";
+        assert("Link not available!");
```
patrickelectric: We should not assert in something that the user can fix, like choosing a different link or something. `assert` should only be used in places where it's not possible to recover.

src/link/link.cpp
```diff
@@ -19,29 +19,28 @@ Link::Link(LinkType linkType, QString name, QObject* parent)
 {
     switch(linkType) {
     case LinkType::None :
-        _abstractLink.reset(new AbstractLink());
+        _abstractLink = new AbstractLink(this);
```
patrickelectric: The class was created without parent. I don't see the point here. This is also more error prone, E.g: Why are you not deleting `_abstractLink`?

patrickelectric: Also, I'm not aware of such thing as as crashing with Qt and smartpointers, should we stay with smartPointers and remove the setParent ? This appears to be the less error prone option IMO.

src/link/seriallink.h
```diff
@@ -80,4 +80,5 @@ class SerialLink : public AbstractLink
 
 private:
     QSerialPort _port;
+    QStringList _availableConnections;
```
patrickelectric: alphabetic order, put available before port.

src/notification/notificationmodel.cpp
```diff
@@ -19,11 +19,13 @@ void NotificationModel::add(QVariant text, QVariant icon, QVariant color, QVaria
     const int line = rowCount();
     beginInsertRows(QModelIndex(), line, line);
     insertRows(line, 1);
-    setData(index(line), time, NotificationModel::Time);
-    setData(index(line), text, NotificationModel::Text);
-    setData(index(line), color, NotificationModel::Color);
-    setData(index(line), severity, NotificationModel::Severity);
-    setData(index(line), icon, NotificationModel::Icon);
+    const QModelIndex idx = index(line);
+    /* This should be just *one* call. */
```
patrickelectric: Is this comment a TODO: ?

src/sensor/hexvalidator.cpp
```diff
@@ -31,7 +31,7 @@ bool HexValidator::check(const QByteArray& bytes)
     }
 
     // Check if it has an odd number of elements
-    if(bytes.size()%2 == 0) {
+    if(bytes.size() & 0x1) {
```
patrickelectric: Don't try to be smarter than the compiler ;)
image

src/sensor/parser-ping.cpp
```diff
@@ -23,12 +23,25 @@ void PingParserExt::parseBuffer(const QByteArray& data)
...
+        case PingParser::ParseState::WAIT_LENGTH_L:
+        case PingParser::ParseState::WAIT_LENGTH_H:
+        case PingParser::ParseState::WAIT_MSG_ID_L:
+        case PingParser::ParseState::WAIT_MSG_ID_H:
+        case PingParser::ParseState::WAIT_SRC_ID:
+        case PingParser::ParseState::WAIT_DST_ID:
+        case PingParser::ParseState::WAIT_PAYLOAD:
+        case PingParser::ParseState::WAIT_CHECKSUM_L:
+        case PingParser::ParseState::WAIT_CHECKSUM_H:
+            return Parser::ParserState::PARSING;
```
patrickelectric: I don't see the point here, the compiler already checks if `==` is a valid enum class and not a int.

src/devicemanager/devicemanager.h
```diff
@@ -30,7 +30,7 @@ class DeviceManager : public QAbstractListModel
      * @return DeviceManager*
      */
     static DeviceManager* self();
-    ~DeviceManager();
+    ~DeviceManager() override;
```
patrickelectric: The commit is not clear. Why is this necessary ? Why override is necessary ? Why are you overriding the desconstrutor. 
Is this overriding the parent deconstruction to avoid any problem related to the singleton nature ?

src/flash/flasher.cpp
```diff
@@ -11,6 +11,9 @@
 
 PING_LOGGING_CATEGORY(FLASH, "ping.flash")
 
+namespace {
+    const QVector<int> validBaudRates = {57600, 115200, 230400};
+}
```
patrickelectric: Please add a space between the namespace and the class constructor.
Why are you moving from privated to namespace ?
Please, be more clear in your commit.

src/flash/flasher.h
```diff
@@ -33,7 +33,7 @@ class Flasher : public QObject
      * @brief Defines flash state
      *
      */
-    enum States {
+    enum class States : char {
```
patrickelectric: Since this is a numeric representation, use int8_t.

src/link/linkconfiguration.h
```diff
@@ -37,7 +37,7 @@ class LinkConfiguration : public QObject
      *
      */
     enum Error {
-        MissingConfiguration, // This can be used in future for warnings and not real errors
+        MissingConfiguration = -1, // This can be used in future for warnings and not real errors
```
patrickelectric: good catch.

src/link/filelink.cpp
```diff
@@ -34,9 +34,8 @@ void FileLink::_writeData(const QByteArray& data)
 
     // This save the data as a structure to deal with the timestamp
     if(_openModeFlag == QIODevice::WriteOnly && _file.isWritable()) {
-        QString time =  QTime::fromMSecsSinceStartOfDay(_timer.elapsed()).toString(_timeFormat);
-        Pack pack{time, data};
-        _inout << pack.time << pack.data;
+        _inout << QTime::fromMSecsSinceStartOfDay(_timer.elapsed()).toString(_timeFormat)
```
patrickelectric: The usage of the structure here is less error prone, you are creating the structure by hand here, and this does not sync automatically with the Pack definition, resulting in a less readable and more error prone approach.

src/sensor/ping.cpp
```diff
@@ -497,7 +497,7 @@ void Ping::checkNewFirmwareInGitHubPayload(const QJsonDocument& jsonDocument)
     float lastVersionAvailable = 0.0;
 
     auto filesPayload = jsonDocument.array();
-    for(const QJsonValue& filePayload : filesPayload) {
+    for(const QJsonValue filePayload : filesPayload) {
```
patrickelectric: I believe that you want to use QJsonValueRef here since you are moving away from reference.

src/devicemanager/devicemanager.cpp
```diff
@@ -95,10 +97,10 @@ void DeviceManager::connectLink(LinkConfiguration* linkConf)
...
     // We could use a single Ping instance, but since we are going to support multiple devices
     // this pointer will hold everything for us
-    if(linkConf->deviceType() == PingDeviceType::PING1D) {
-        _primarySensor.reset(new Ping());
-    } else {
-        _primarySensor.reset(new Ping360());
+    switch(linkConf->deviceType()) {
+    case PingDeviceType::PING1D : _primarySensor.reset(new Ping()); break;
+    case PingDeviceType::PING360 : _primarySensor.reset(new Ping360()); break;
+    case PingDeviceType::UNKNOWN : assert("Should never happen"); break;
```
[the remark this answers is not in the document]
tcanabrava: We can use a Ping1D fallback if you prefer. but if the type goes tehre as unknown it's actually a programming error and aborting at least gives us the backtrace to discover what's wrong.

src/flash/flasher.cpp
tcanabrava: I'll rework the commit messages today. anything that's private / static should go into the cpp's unammed namespace, this will reduce the compilation time for the program.

src/link/filelink.cpp
tcanabrava: ok, I'll ignore this patch.

src/sensor/ping.cpp
tcanabrava: Probably. I actually would use nlohmann json for this, seems saner. but right now we are not depending on it and I don't want to add new dependencies.

src/flash/flasher.cpp
tcanabrava: this is really bad to read :)

patrickelectric:
@tcanabrava 	please rebase and fix CI when possible.

src/link/abstractlink.h
```diff
@@ -225,7 +225,7 @@ class AbstractLink : public QObject
     const AbstractLink& operator=(const AbstractLink& other);
 
     Q_PROPERTY(qint64 byteSize READ byteSize NOTIFY byteSizeChanged)
-    Q_PROPERTY(LinkConfiguration* configuration READ configuration CONSTANT)
+    Q_PROPERTY(const LinkConfiguration* configuration READ configuration CONSTANT)
```
patrickelectric: I believe that this is breaking PingStatus widget:
image

tcanabrava: hm.

tcanabrava:
closing the review: too many commits that are unrelated to eachother makes this hard to test. I'll open pull requests that are more coherent.

PR #563, opened 2019-07-19
Ping360 parameters

### LINKED ISSUE #546: Changing scan range does not affect sonar settings
When the scan range (length_mm) is changed by the user, the axis labels change, but the sonar settings themselves do not change. This behavior can be readily observed with https://github.com/bluerobotics/ping-viewer/pull/545.

### LINKED ISSUE #550: ping360 visualizer range labels are out of sync with actual device settings
The range must be recalculated/emitted (`scanLengthChanged`) every time `sample_period` or `number_of_points` changes (when we receive `device_data` message from the device)

jaxxzer:
fix #546 
fix #550

### DIFF
--- a/src/sensor/ping360.cpp
+++ b/src/sensor/ping360.cpp
@@ -151,6 +151,32 @@ void Ping360::handleMessage(const ping_message& msg)
         break;
     }
 
+    case CommonId::NACK: {
+        const common_nack nack(msg);
+        if (nack.nacked_id() == Ping360Id::TRANSDUCER) {
+            qCWarning(PING_PROTOCOL_PING360) << "transducer control was NACKED, reverting to default settings";
+
+            _gain_setting = _firmwareDefaultGainSetting;
+            _transmit_duration = _firmwareDefaultTransmitDuration;
+            _sample_period = _firmwareDefaultSamplePeriod;
+            _transmit_frequency = _firmwareDefaultTransmitFrequency;
+            _num_points = _firmwareDefaultNumberOfSamples;
+
+            // request another transmission
+            requestNextProfile();
+
+            // restart timer
+            _timeoutProfileMessage.start();
+
+            emit gainSettingChanged();
+            emit transmitDurationChanged();
+            emit samplePeriodChanged();
+            emit transmitFrequencyChanged();
+            emit numberOfPointsChanged();
+        }
+        break;
+    }
+
     default:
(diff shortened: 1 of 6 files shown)


src/sensor/ping360.cpp
```diff
@@ -151,6 +151,32 @@ void Ping360::handleMessage(const ping_message& msg)
         break;
     }
 
+    case CommonId::NACK: {
+        const common_nack nack(msg);
+        if (nack.nacked_id() == Ping360Id::TRANSDUCER) {
+            qCWarning(PING_PROTOCOL_PING360) << "transducer control was NACKED, reverting to default settings";
+
```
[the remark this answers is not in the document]
jaxxzer: Yes. We will improve on this behavior in the future. Presently, this is an improvement on the current behavior.

jaxxzer:
> The range labels need to be changed to handle meters.

please clarify.

> If I change the transmit duration to 1000, sometimes the range changes to a very small one.

This is unreleated to this pull request.

> The range now starts at 536 meters, that looks wrong.

This is an existing issue that will be fixed in another pull request.

> Are we not using the formula for transmit_duration?

No, we have a lot of other problems to be addressed first.

Williangalvani:
> > The range labels need to be changed to handle meters.
> 
> please clarify.

There is a 1000x factor error here:
Screenshot from 2019-07-19 16-44-19

> This is unrelated to this pull request. You probably entered an invalid configuration.

So the sensor refused it? shouldn't my settings have been set to default, then?

jaxxzer:
> So the sensor refused it? shouldn't my settings have been set to default, then?

It's unclear without information about all of the other settings. We will make the settings adjustment more robust later. In master, it does not work **at all**. This pull request fixes bugs so that they do work.

jaxxzer:
> There is a 1000x factor error here:
> Screenshot from 2019-07-19 16-44-19

Fixed ty

src/sensor/ping360.cpp
```diff
@@ -151,6 +145,32 @@ void Ping360::handleMessage(const ping_message& msg)
...
+            requestNextProfile();
+
+            // restart timer
+            _timeoutProfileMessage.start();
+
+            emit gainSettingChanged();
+            emit transmitDurationChanged();
+            emit samplePeriodChanged();
+            emit transmitFrequencyChanged();
+            emit numberOfPointsChanged();
```
Williangalvani: `emit rangeChanged();` here to solves the issue I have messing with the transmit duration.

jaxxzer: I know ;)

PR #564, opened 2019-07-19
Small fixes

Williangalvani:
This prevents invalid IP addresses and Ports. I had some crashes because of missing validation.

Also moves the Tooltip down closer to the device button, and fixes the transmit_duration unit.

### DIFF
--- a/qml/ConnectionMenu.qml
+++ b/qml/ConnectionMenu.qml
@@ -175,7 +175,7 @@ Item {
 
                 Text {
                     text: "UDP Host/Port:"
-                    color: Material.primary
+                    color: udpIp.isValid ? Material.primary : Material.color(Material.Error)
                 }
 
                 PingTextField {
@@ -183,12 +183,16 @@ Item {
                     text: "192.168.2.2"
                     Layout.columnSpan:  2
                     Layout.fillWidth: true
+                    property bool isValid:  // Check for valid IPV4 (0.0.0.0, 192.168.0.2)
+                                            text.match(/^(?!\.)((^|\.)([1-9]?\d|1\d\d|2(5[0-5]|[0-4]\d))){4}$/gm)
+                                            // Check for valid host names (potato.com, smashed.potato.com.br, great-potato-of-doom.com)
+                                            || text.match(/^(([a-zA-Z0-9-]+\.){0,5}[a-zA-Z0-9-][a-zA-Z0-9-]+\.[a-zA-Z]{2,63}?)$/gm)
+                                            // Check for a valid host name (localhost, raspberrypi, potato)
+                                            || text.match(/^[a-zA-Z0-9-]*$/gm)
                     onEditingFinished: {
-                        if (text === "0.0.0.0" || text === "localhost") {
-                            text = "127.0.0.1"
+                        if (isValid) {
+                            connect(AbstractLinkNamespace.Udp, text, udpPort.text)
                         }
-
-                        connect(AbstractLinkNamespace.Udp, text, udpPort.text)
                     }
                 }
 
@@ -197,6 +201,10 @@ Item {
                     text: "1234"
                     Layout.columnSpan:  2
                     Layout.fillWidth: true
+                    validator: IntValidator {
+                        bottom: 1
+                        top: 65535
+                    }
                     onEditingFinished: {
                         connect(AbstractLinkNamespace.Udp, udpIp.text, text)
                     }
--- a/qml/DeviceManagerViewer.qml
+++ b/qml/DeviceManagerViewer.qml
@@ -148,6 +148,8 @@ PingPopup {
                             root.close()
                         }
                         ToolTip {
+                            x: parent.mouseX
+                            y: parent.mouseY - height - 5
                             visible: parent.containsMouse
                             text: "Connect."
                         }
--- a/src/link/udplink.cpp
+++ b/src/link/udplink.cpp
@@ -32,7 +32,7 @@ bool UDPLink::setConfiguration(const LinkConfiguration& linkConfiguration)
 
     setName(linkConfiguration.name());
 
-    _hostAddress = QHostAddress(linkConfiguration.args()->at(0));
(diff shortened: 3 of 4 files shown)


qml/DeviceManagerViewer.qml
```diff
@@ -148,6 +148,8 @@ PingPopup {
                             root.close()
                         }
                         ToolTip {
+                            // Position tooltip slightly above parent.
+                            y: parent.y - height - 5
```
patrickelectric: Use margin if possible, avoid magic numbers.

qml/ConnectionMenu.qml
```diff
@@ -183,6 +183,8 @@ Item {
                     text: "192.168.2.2"
                     Layout.columnSpan:  2
                     Layout.fillWidth: true
+                    // This regex matches IPV4 Addresses. From https://stackoverflow.com/a/50524903
+                    validator: RegExpValidator { regExp: /^(?!\.)((^|\.)([1-9]?\d|1\d\d|2(5[0-5]|[0-4]\d))){4}$/gm }
```
jaxxzer: "localhost" should be allowed.

qml/Ping360ControlPanel.qml
```diff
@@ -71,7 +71,7 @@ Item {
                 }
 
                 PingTextField {
-                    title: "Transmit Duration (ms):"
+                    title: "Transmit Duration (us):"
```
jaxxzer: stale, ty/sorry

qml/ConnectionMenu.qml
Williangalvani: Good point. any other domain should be valid too (raspberrypi.local etc)

patrickelectric:
@Williangalvani 	please update the PR.

qml/DeviceManagerViewer.qml
Williangalvani: I tried, bu couldn't get anything better to work.

qml/ConnectionMenu.qml
Williangalvani: updated the regex so it handles both options nicely.

Williangalvani:
Updated. The regex is now much longer than I would like. But I spent far too long trying to break it down and am officially giving up. Help and suggestions are welcome.

qml/ConnectionMenu.qml
```diff
@@ -175,20 +175,23 @@ Item {
...
 
                 PingTextField {
                     id: udpIp
                     text: "192.168.2.2"
                     Layout.columnSpan:  2
                     Layout.fillWidth: true
+                    // This regex matches IPV4 Addresses and valid domain names.
+                    // See https://www.regextester.com/index.php?fam=110631
+                    // TODO: try to simplify/split this
+                    property bool isValid: text.match(/^(((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)|(([a-zA-Z0-9-]+\.){0,5}[a-zA-Z0-9-][a-zA-Z0-9-]+\.[a-zA-Z]{2,63}?)|[a-zA-Z0-9-]+)$/gm)
```
patrickelectric: Is this better than a if and 2~3 checks ?

patrickelectric: Just to make sure, is this part equal to:
```
var intput;
// Check for valid ipv4 IP address
regex = "^(?!\.)((^|\.)([1-9]?\d|1\d\d|2(5[0-5]|[0-4]\d))){4}$"
if(regex.check(input)) { output = intput; return; }

// check if it's a normal domain
if(input.split('.') <= 2) { output = intput; return; }
```

You could also break the `or` of this same regex in multiple regex to allow a better readability and maintainability.
```js
property bool isValid:
                        // Check for valid IPV4 (0.0.0.0, 192.168.0.2)
                        text.match(/^(?!\.)((^|\.)([1-9]?\d|1\d\d|2(5[0-5]|[0-4]\d))){4}$/gm)
                        // Check for valid host names (potato.com, smashed.potato.com.br, great-potato-of-doom.com)
                        || text.match(/^(([a-zA-Z0-9-]+\.){0,5}[a-zA-Z0-9-][a-zA-Z0-9-]+\.[a-zA-Z]{2,63}?)$/gm)
                        // Check for a valid host name (localhost, raspberrypi, potato)
                        || text.match(/^[a-zA-Z0-9-]*$/gm)
```

qml/DeviceManagerViewer.qml
patrickelectric: You can use the parent to use the mouse position and the height to make it more friendly and with a better position.
```
x: parent.mouseX
y: parent.mouseY - height - 5
```
image
deepin-screen-recorder_Select area_20190801102824

src/link/udplink.cpp
```diff
@@ -32,7 +32,7 @@ bool UDPLink::setConfiguration(const LinkConfiguration& linkConfiguration)
 
     setName(linkConfiguration.name());
 
-    _hostAddress = QHostAddress(linkConfiguration.args()->at(0));
+    QString _hostAddress = linkConfiguration.args()->at(0);
```
patrickelectric: you are shadowing the already existent private variable with the same name.

qml/ConnectionMenu.qml
```diff
@@ -175,20 +175,25 @@ Item {
 
                 Text {
                     text: "UDP Host/Port:"
-                    color: Material.primary
+                    color: udpIp.isValid ? Material.primary : Material.color(Material.Red)
```
patrickelectric:
```suggestion
					color: udpIp.isValid ? Material.primary : Material.color(Material.Error)
```

Williangalvani: Nice. I was actually looking for that. Where can I find these colors?

patrickelectric: https://material.io/design/color/#color-theme-creation

PR #584, opened 2019-07-23
Ping speedups

tcanabrava:
Draft, and I tried to make this to be based in my last PR, that's the reason that there are too many diffs, but ignore all of them untill the last 9: only they matter for this pull request. Github has no way for me to base a pull request on code that's still in review.

Look from 
Fix 4 byte hole in class 
to 
Pre-calculate multiplication


tcanabrava:
closing this because of the amount of noise. I'll open a new PR way smaller than this one with the same behavior.

PR #587, opened 2019-07-24
Ping360 dev

### LINKED ISSUE #568: Firmware version is not displayed correctly
On ping360: 
Screenshot_20190722_153803

### LINKED ISSUE #570: ping360 frequency should default to 750kHz
per https://github.com/bluerobotics/ping-viewer/pull/565#issuecomment-513884228

### LINKED ISSUE #574: ping360: use maximum number of samples
Right now we are using 1024 points, we should use the maximum resolution possible.

### LINKED ISSUE #575: Allow range < 1.54
To allow short ranges, we will need to reduce the number of samples, or truncate the data when we plot it.

jaxxzer:
fix #575 
fix #574 
fix #570 
fix #568

### DIFF
--- a/qml/Ping360ControlPanel.qml
+++ b/qml/Ping360ControlPanel.qml
@@ -68,7 +68,6 @@ Item {
                 value: ping.transmit_frequency
                 control.from: 500
                 control.to: 1000
-                control.stepSize: 20
                 control.onMoved: ping.transmit_frequency = control.value
             }
             PingSlider {
--- a/qml/Ping360ControlPanel.qml
+++ b/qml/Ping360ControlPanel.qml
@@ -94,8 +94,6 @@ Item {
                 control.to: 359
                 control.onMoved: ping.angle_offset = control.value
             }
-
-
         }
     }
 }
--- a/src/sensor/ping360.h
+++ b/src/sensor/ping360.h
@@ -358,12 +358,15 @@ private:
     static const uint16_t _firmwareDefaultTransmitFrequency = 740;
     static const uint16_t _firmwareDefaultNumberOfSamples = 1024;
 
+    // The default transmit frequency to operate with
+    static const uint16_t _viewerDefaultTransmitFrequency = 750;
+
     uint16_t _angle = _firmwareDefaultAngle;
     uint16_t _transmit_duration = _firmwareDefaultTransmitDuration;
     uint32_t _gain_setting = _firmwareDefaultGainSetting;
     uint16_t _num_points = _firmwareDefaultNumberOfSamples;
     uint16_t _sample_period = _firmwareDefaultSamplePeriod;
-    uint16_t _transmit_frequency = _firmwareDefaultTransmitFrequency;
+    uint16_t _transmit_frequency = _viewerDefaultTransmitFrequency;
     QVector<double> _data;
 ///@}
 
--- a/src/sensor/ping360.h
+++ b/src/sensor/ping360.h
@@ -157,7 +157,15 @@ public:
     void set_range(uint newRange)
     {
         if(newRange != range()) {
+            _num_points = _firmwareMaxNumberOfPoints;
             _sample_period = ceil(2*newRange/(_num_points*_speed_of_sound*_samplePeriodTickDuration));
+
+            // reduce _sample period until we are within operational parameters
+            // maximize the number of points
+            while(_sample_period < _firmwareMinSamplePeriod) {
+                _num_points--;
+                _sample_period = ceil(2*newRange/(_num_points*_speed_of_sound*_samplePeriodTickDuration));
+            }
             emit rangeChanged();
         }
     }
@@ -352,6 +360,7 @@ private:
 
     // firmware constants
     static const uint16_t _firmwareMaxNumberOfPoints = 1200;
+    static const uint16_t _firmwareMinSamplePeriod = 80;
     // The firmware defaults at boot
     static const uint8_t _firmwareDefaultGainSetting = 0;
     static const uint16_t _firmwareDefaultAngle = 0;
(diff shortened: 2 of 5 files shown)


src/sensor/ping360.h
```diff
@@ -358,12 +358,15 @@ class Ping360 : public PingSensor
     static const uint16_t _firmwareDefaultTransmitFrequency = 740;
     static const uint16_t _firmwareDefaultNumberOfSamples = 1024;
 
+    // The default transmit frequency to operate with
+    static const uint16_t _viewerDefaultTransmitFrequency = 750;
```
patrickelectric: what is the difference between this variable and _firmwareDefaultTransmitFrequency ?

src/sensor/ping360.h
```diff
@@ -157,7 +157,15 @@ class Ping360 : public PingSensor
     void set_range(uint newRange)
     {
         if(newRange != range()) {
+            _num_points = _firmwareMaxNumberOfPoints;
```
patrickelectric: so it's not possible to control range and points individually ? That's not a good thing since we are not allowing our users to run minimal distance with minimal number of points to increase the fps.

src/sensor/ping360.h
jaxxzer: The firmware uses 740 khz by default.

src/sensor/ping360.h
jaxxzer: > so it's not possible to control range and points individually ?

It is already this way in master, this pull request does not affect that. If we require direct control over number of points, we need to open a new issue to add it independently of this pull request, and we can have the option of automatic maximizing like we have implemented in master already. Very short distance scanning <2m does not represent a significant portion of use cases, and the performance gain will be marginal (the serial time saved by changing number of points is insignificant), so this is low priority in my opinon.

I think it is more useful for our application to always use the maximum resolution to provide the user with the highest detail achievable with the hardware. Range and number of points are coupled by an equation. It makes the most sense to me to allow the user to adjust only one. This makes the application easy to use.

jaxxzer:
@patrickelectric let me know anything to change.

qml/Ping360ControlPanel.qml
```diff
@@ -68,7 +68,7 @@ Item {
                 value: ping.transmit_frequency
                 control.from: 500
                 control.to: 1000
-                control.stepSize: 20
+                control.stepSize: 10
```
patrickelectric: Why not steps of 1 ? Since we merged the wheel control this will makes the user input more precise. There's some users that really want to use precise numbers.

src/sensor/ping360.h
patrickelectric: My points is, where `_firmwareDefaultTransmitFrequency` is being used ?

qml/Ping360ControlPanel.qml
jaxxzer: Ok, will do

jaxxzer:
@patrickelectric I added also one commit: 3cf817e

src/sensor/ping360.h
jaxxzer: It is not, but it is a useful constant to have for reference.

src/sensor/pingsensor.h
```diff
@@ -197,19 +197,19 @@ class PingSensor : public Sensor
...
-    uint8_t _firmware_version_major;
-    uint8_t _firmware_version_minor;
-    uint8_t _firmware_version_patch;
-    int _lostMessages;
-    QString _nack_msg;
-    uint8_t _protocol_version_major;
-    uint8_t _protocol_version_minor;
-    uint8_t _protocol_version_patch;
-    uint8_t _srcId;
+    QString _ascii_text = QString();
```
patrickelectric: QStrings are not necessary here, the default constructors do that, besides this forces the unnecessary creation and move of an empty string.

jaxxzer: What about the zero for integers? Is this different?

I thought maybe it is a problem with the dynamic creation of the object.

> this forces the unnecessary creation and move of an empty string.

I think @tcanabrava has mentioned QString is smart enough that this doesn't happen.

tcanabrava: it's different: *if* QStrings where a class without default constructor, then nothing would happen as the compiler wouldn't know what to do and garbage would be inside the values. But QString is a class with a default constructor that creates an empty string. int, on the other hand, has no default value initializer = you can call int a{0}; to default initialize it to 0, but if you just do a `int i;` there's gargage in memory.

jaxxzer: perfect :100:

src/sensor/pingsensor.h
```diff
@@ -198,18 +198,18 @@ class PingSensor : public Sensor
...
 
     QString _ascii_text;
-    uint8_t _device_revision;
-    uint8_t _device_type;
-    uint8_t _dstId;
-    uint8_t _firmware_version_major;
-    uint8_t _firmware_version_minor;
-    uint8_t _firmware_version_patch;
-    int _lostMessages;
+    uint8_t _device_revision {0};
```
patrickelectric: Usually the uniform initialization is done with the braces next to the variable and not with a space between it, thing it like a constructor call.  
https://en.cppreference.com/w/cpp/language/list_initialization

jaxxzer: aha

jaxxzer:
@patrickelectric finished.

src/sensor/ping360.h
patrickelectric: Note: Should we ask "Ping360 developer" to set 750 as default ?

patrickelectric:
@jaxxzer 	there is the re-review request button next to the reviewers tool.

PR #590, opened 2019-07-24
[WIP] PingSlider: Do not limit max size of elements 

patrickelectric:
Elements size should be specified by the control itself but for the element content.
Any necessary tweak around that is a specific usage of the parent component of PingSlider.

### DIFF
--- a/qml/PingSlider.qml
+++ b/qml/PingSlider.qml
@@ -6,8 +6,8 @@ import QtQuick.Layouts 1.3
 RowLayout {
     id: root
 
-    property var text
-    property var value
+    property string text
+    property string value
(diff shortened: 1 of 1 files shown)


jaxxzer:
I understand the way that I fixed the width is technically incorrect. I thought it gave a better aesthetic. Is it possible to get the text fields and sliders aligned like this? 
image


Yours looks like this:
Screenshot_20190724_170024

tcanabrava:
I like the work that me / jaxxed did with the new combos and I feel that when they are used this is actually a nice change.

tcanabrava:
I believe here the way to go is to limit by percentage.

PR #573, opened 2019-07-22
qml: Add new Qml entity: ComboSlider

tcanabrava:
fix #571
fix #594

PR #602, opened 2019-07-26
Ping360 transmit duration

### LINKED ISSUE #585: Add automatic transmit_duration adjustment

### LINKED ISSUE #599: ping360: constrain transmit_duration values to valid values
The valid range of `transmit_duration` depends on the other settings. We should add some logic to avoid invalid configurations.

jaxxzer:
This is based on #587 

fix #599 
fix #585 

The auto transmit duration can be made optional when when #573 is fixed and merged


src/sensor/ping360.h
```diff
@@ -319,6 +327,58 @@ class Ping360 : public PingSensor
...
+    }
+    Q_PROPERTY(int transmitDurationMax READ transmitDurationMax NOTIFY transmitDurationMaxChanged)
+
+    /**
+     * @brief automatic transmit duration adjustment
+     * @param automatic true to adjust the transmit duration automatically when the range is adjusted
+     */
+    void setAutoTransmitDuration(bool automatic)
+    {
+        if (_autoTransmitDuration != automatic) {
```
patrickelectric: to avoid multiple and future indentation levels, try to use == return;

src/sensor/ping360.h
```diff
@@ -319,6 +327,58 @@ class Ping360 : public PingSensor
...
+    Q_PROPERTY(bool autoTransmitDuration READ autoTransmitDuration WRITE setAutoTransmitDuration NOTIFY
+               autoTransmitDurationChanged)
+
+    /**
+     * @brief adjust the transmit duration according to automatic mode, and current configuration
+     */
+    void adjustTransmitDuration()
+    {
+        if (_autoTransmitDuration) {
+            int autoDuration = round(5.1 + 2.833*(range() - 2));
```
patrickelectric: add note about magic equation that was reverse engineered

src/sensor/ping360.h
```diff
@@ -358,8 +420,11 @@ class Ping360 : public PingSensor
      */
 ///@{
 
+    // TODO: static members should be defined in .cpp
     // firmware constants
     static const uint16_t _firmwareMaxNumberOfPoints = 1200;
+    const uint16_t _firmwareMaxTransmitDuration = 500;
+    const uint16_t _firmwareMinTransmitDuration = 5;
```
patrickelectric: static const

src/sensor/ping360.h
patrickelectric: We should use the new equation that was shared with us.

src/sensor/ping360.h
tcanabrava: if this is private, I would move it to the .cpp file and use constexpr in an unamed namespace.

jaxxzer: Check the TODO added above. This is an existing issue.

jaxxzer: I will move them to anonymous namespace in cpp.

qml/Ping360ControlPanel.qml
```diff
@@ -59,7 +59,7 @@ Item {
                 text: "Transmit Duration (μs)"
                 value: ping.transmit_duration
                 control.from: 5
```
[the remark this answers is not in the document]
jaxxzer: We should have constants for everything. I am working toward that here: https://github.com/bluerobotics/ping-protocol/pull/161

src/sensor/ping360.h
jaxxzer: we can't do anonymous namespace because we have code using the members in .h. I moved the definitions to .cpp, they are still class members, `static const`.

jaxxzer:
@patrickelectric done

PR #605, opened 2019-07-26
protocoldetector: Break any existing loop if _activate is false

patrickelectric:

### DIFF
--- a/src/sensor/protocoldetector.cpp
+++ b/src/sensor/protocoldetector.cpp
@@ -51,10 +51,14 @@ void ProtocolDetector::doScan()
 {
     _active = true;
     LinkConfiguration linkConf;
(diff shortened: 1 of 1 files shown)


jaxxzer:
Is this fixing some bug or behavior?

patrickelectric:
It avoids unnecessary debug warnings/errors about ports that are used by the sensor class. Also fix a rare error that fails the sensor connection in a little small time interval that the protocoldetector is scanning the same port.

jaxxzer:
> It avoids unnecessary debug warnings/errors about ports that are used by the sensor class. Also fix a rare error that fails the sensor connection in a little small time interval that the protocoldetector is scanning the same port.

This is a perfect commit body ;)

PR #606, opened 2019-07-26
protocoldetector: scan for high baudrates, drop 9600 for now

### LINKED ISSUE #558: ping360 - support connections up to 2Mbps
Provide a range of serial baudrates to work with a range of tether lengths. The maximum baudrate is 2Mbps.

In the future, it will be awesome to autobaud to the highest error-free baudrate.

jaxxzer:
This allows connecting to ping360 at 2M. There is some problem with the firmware in moving from 9600 to 2M baudrate or something, some combinations work, some do not. We need to investigate the exact problem more.

I think we will be able to work it out to add 9600 back in the future. Otherwise, it should be possible to connect a ping1d@9600 with manual connection.

fix #558

### DIFF
--- a/src/sensor/protocoldetector.cpp
+++ b/src/sensor/protocoldetector.cpp
@@ -100,9 +100,9 @@ QVector<LinkConfiguration> ProtocolDetector::updateLinkConfigurations(QVector<Li
         }
 
         // Add valid port and baudrate
-        // Ping360 can't handle 9600 requests with 115200 request in a sort time priod
+        // Ping360 can't handle 9600 requests with 2000000 request in a short time priod, look for bug in firmware
         // TODO: Fix Ping360 is not possible, we should drop 9600 checks if 115200 returns fine
-        for(auto baud : {115200, 9600}) {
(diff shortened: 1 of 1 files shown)


patrickelectric:
Why not deal with it in the sensor level ? We can't cover all possible baud rates in protocol detector, this can also result in  weird behaviors like the one that you are talking about 9600.

We should stay with at least this two basic baud rates that ensure that any device can be detected, allowing slow software serial from arduinos and normal sensors with 115200 baudrate, I'm also willing to move from 115200 to something better inside the range of 230400-1M. 

Also, since the sensor needs to deal with auto baud rate problems and baud rate modifications to allow a better performance and noise channels, this points it that such behavior needs to be done in sensor level and not in protocol detector.

jaxxzer:
I am open to reworking the implementation. With the current implementation, it takes one commit (here) to fix our outstanding issues.

If you have a plan for reworking the baudrate selection at an architectural level, please open a proposal in a new issue.

Meanwhile, we should strive to deliver a better working application to our engineers for testing, and move more quickly on functional changes.

patrickelectric:
Check #614

PR #613, opened 2019-07-26
Fix polar misalignment exceeds sector limits 

patrickelectric:


src/waterfall/polarplot.cpp
```diff
@@ -60,20 +60,21 @@ void PolarPlot::setImage(const QImage &image)
     setImplicitHeight(image.height());
 }
 
-void PolarPlot::draw(const QVector<double>& points, float angle, float initPoint, float length, float angleGrad)
+void PolarPlot::draw(const QVector<double>& points, float angle, float initPoint, float length, float angleGrad,
+                     float sectorSize)
```
tcanabrava: this is startig to feel like too many variables, maybe pack them in a struct?

patrickelectric: Somehow github lost my comment...
It's not possible to send structures between QML and C++, this is only possible via Qt Objects.
Creating a single Qt Object just to be used in a single function that is used in a single qml file and sensor does not make sense to me right now.

This number of variables are something that is related to the sensor and to the visualization settings that we provide, I would agree that we could split this function in more functions to allow a better readability of this same class.

tcanabrava:
Then +1

Em ter, 30 de jul de 2019 às 14:15, Patrick José Pereira <

Issue #596, opened 2019-07-25
ping360: hide advanced controls by default


jaxxzer:
We can hide the advanced settings beneath the white line, like we do with ping1dcontrolpanel.

PR #616, opened 2019-07-29
Ping360ControlPanel: Add advanced settings button

patrickelectric:
Fix [#596](https://github.com/bluerobotics/ping-viewer/issues/596)

### DIFF
--- a/qml/MainPage.qml
+++ b/qml/MainPage.qml
@@ -49,7 +49,7 @@ Item {
 
         PingItem {
             id: settingsMenu
-            marginMult: 1
+            margin: 0
             icon: StyleManager.settingsIcon()
             item: ColumnLayout {
                 spacing: 0
--- a/qml/Ping360ControlPanel.qml
+++ b/qml/Ping360ControlPanel.qml
@@ -91,7 +91,6 @@ Item {
                 visible: advancedSettingsButton.selected
                 PingSlider {
                     Layout.fillWidth: true
-                    // TODO: proper symbol for us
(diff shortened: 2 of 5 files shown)


tcanabrava:
+1 for what willian said

patrickelectric:
> The whole thing shifts down a little when the advanced options are toggled:

This is something related to PingItem check #617. This is fixed in the last version of this PR.

> I think the white line should be hidden too.

Removed to work like Ping1D.

> That empty space to the left of the icon is bothering me. I'm not sure of what to do to fix it. Maybe write "Advanced settings" in the empty space?

This is because we don't have some switches or other buttons to use the same, I'm aiming to use the same thing as we have already in Ping1D to maintain a consistent UI. I have added the firmware update button but with a disabled color to make you happier. 

Fixed.

Williangalvani:
That is a lot better <3.
Can this be easily fixed?
```
default[Warning]: Ping360ControlPanel.qml(57) qrc:/Ping360ControlPanel.qml:57:13: QML RowLayout: Detected anchors on an item that is managed by a layout. This is undefined behavior; use Layout.alignment instead.
```

Are these margins ok for hi-dpi displays too?

patrickelectric:
> Can this be easily fixed?

Done.



> Are these margins ok for hi-dpi displays too?

Yes, they work as a normal margin, hidpi problem are usually something related to images or Qt configuration, this was fixed with 0a07b3458253eb8a58b3f2d2c6e2bc2266a6f349

But just to make sure:
normal screen:
image

macbook:
image

PR #618, opened 2019-07-30
Add frequency information about profile messages 

### LINKED ISSUE #580: FPS and/or full circle sweep time displayed in debug window
## Summary
I think it would be valuable to calculate and print the frames per second and/or the total 360 sweep time in the debug window. Not super critical but useful information.

patrickelectric:
Fix #580 

deepin-screen-recorder_Select area_20190730095809

### DIFF
--- a/qml/Ping360Status.qml
+++ b/qml/Ping360Status.qml
@@ -30,6 +30,7 @@ Item {
                     "Range (m): " + ping.range.toFixed(2),
                     "Sample period (μs): " + ping.sample_period,
                     "Number of samples (#): " + ping.number_of_points,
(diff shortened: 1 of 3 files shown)


patrickelectric:
That's that indent behavior, this is a debug widget and we want to see this fluctuations around the messages/seconds. As we can see from your gif that something should be done to improve the network layers to allow a better and less variable frequency between messages, such information could not be possible to detect with the 1s timer.

Williangalvani:
I think in this case the total throughput is equally important, as we want the fastest scanning speed possible. But I agree that reducing jitter should improve that.

PR #627, opened 2019-07-31
Improve manual connection

### LINKED ISSUE #612: devicemanager: fix manual connection
Manual connection is apparently not working.

patrickelectric:
Fix #612
Needs: https://github.com/bluerobotics/ping-cpp/pull/8

### DIFF
--- a/qml/ConnectionMenu.qml
+++ b/qml/ConnectionMenu.qml
@@ -5,6 +5,7 @@ import QtQuick.Layouts 1.3
 
 import AbstractLinkNamespace 1.0
 import DeviceManager 1.0
+import PingEnumNamespace 1.0
 import SettingsManager 1.0
 import Util 1.0
 
@@ -12,23 +13,6 @@ Item {
     id: root
     property var ping: null
     property var serialPortList: null
-
-    function connect(connectionTypeEnum) {
-        // Only connect from user input
-        if(!parent.visible) {
-            return
-        }
-
-        // Do not connect if no valid type or input
-        if(connectionTypeEnum < AbstractLinkNamespace.None || connectionTypeEnum >= AbstractLinkNamespace.Last) {
-            print("The connection configuration type is not valid!")
-            return
-        }
-
-        // Transform arguments to Array and remove connectionTypeEnum from it
-        var nextArgs = Array.prototype.slice.call(arguments).slice(1)
-        DeviceManager.connectLinkDirectly(connectionTypeEnum, nextArgs)
-    }
 
     Connections {
         target: ping
@@ -68,7 +52,6 @@ Item {
     GroupBox {
         id: connectionLayout
         title: "Connection configuration"
-        enabled: true
         // Hack
         label.x: width/2 - label.contentWidth/2
         width: parent.width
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -67,6 +70,11 @@ int main(int argc, char *argv[])
         "AbstractLinkNamespace", 1, 0, "AbstractLinkNamespace", "This is another enum."
     );
 
+    qmlRegisterUncreatableMetaObject(
+        PingEnumNamespace::staticMetaObject,
+        "PingEnumNamespace", 1, 0, "PingEnumNamespace", "This is another enum."
```
tcanabrava: "this is another enum" is a bad description. This description is used thru IDE's to give us hints about what the code is.

qml/ConnectionMenu.qml
```diff
@@ -81,39 +82,54 @@ Item {
...
+                text: "Device:"
                 enabled: true
                 color: Material.primary
             }
 
             ComboBox {
+                id: deviceCB
+                Layout.fillWidth: true
+                textRole: "name"
+                //TODO: This model should be provided by someone else and not to be hardcoded here
```
tcanabrava: you can perhaps use the QMetaType information of the menu to retrieve the values, that are generated by Q_ENUM to fill a list model. but that can be in another patch.

qml/ConnectionMenu.qml
```diff
@@ -81,39 +82,54 @@ Item {
...
+                    ListElement {
+                        name: "Ping360"
+                        deviceId: PingEnumNamespace.PingDeviceType.PING360
+                    }
+                }
+            }
+
+            Text {
+                text: "Communication:"
+                enabled: true
```
tcanabrava: why "text" needs to be enabled?

qml/ConnectionMenu.qml
```diff
@@ -197,9 +203,29 @@ Item {
...
-                    onEditingFinished: {
-                        connect(AbstractLinkNamespace.Udp, udpIp.text, text)
+                }
+            }
+
+            Button {
+                text: "Connect"
+                Layout.fillWidth: true
+                Layout.columnSpan: 5
+                onClicked: {
```
tcanabrava: maybe this funcitons should go to c++ ?

qml/ConnectionMenu.qml
```diff
@@ -197,9 +203,29 @@ Item {
                     text: "1234"
                     Layout.columnSpan:  2
                     Layout.fillWidth: true
-                    onEditingFinished: {
-                        connect(AbstractLinkNamespace.Udp, udpIp.text, text)
```
tcanabrava: unrelated to the patch, but 'connect' is a very bad name for anything that deals with Qt.

qml/ConnectionMenu.qml
patrickelectric: The problem here is not the enum, but the map between string and enum.

qml/ConnectionMenu.qml
patrickelectric: There is a C++ function in the end to do the connection, this function only gets the necessary information for each QML control.

qml/ConnectionMenu.qml
patrickelectric: Function removed.

patrickelectric:
@Williangalvani 	please test when possible.

qml/ConnectionMenu.qml
```diff
@@ -205,9 +192,29 @@ Item {
...
                         top: 65535
                     }
-                    onEditingFinished: {
-                        connect(AbstractLinkNamespace.Udp, udpIp.text, text)
+                }
+            }
+
+            Button {
+                text: "Connect"
+                Layout.fillWidth: true
```
Williangalvani: Can we do something like this to block connecting to invalid addreses?
```suggestion
                Layout.fillWidth: true
                // Disable if using an invalid UDP address
                enabled: conntype.currentIndex === 0 || udpIp.isValid
```
Can we get something like this to forbid connections to bad addresses?

PR #629, opened 2019-07-31
settingsmanagerhelper: dont call QVariant.value<bool> or <int>

tcanabrava:
Use .value only for unknown types. Currently we are only using
bool and int that triggered the warning. The fix is simple: a template
function will be swapped in compile time for the actuall call.

### DIFF
--- a/src/settings/settingsmanagerhelper.h
+++ b/src/settings/settingsmanagerhelper.h
@@ -2,6 +2,32 @@
  * @brief This header defines some macros to help SettingsManager development
  */
 
+#include <QVariant>
+
+/**
+ *@brief call the correct method for Qvariants that have bool values
+ */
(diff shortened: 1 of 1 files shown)


src/settings/settingsmanagerhelper.h
```diff
@@ -2,6 +2,20 @@
...
+    return variant.toBool();
+}
+
+template<int> int value(const QVariant& variant) {
+    return variant.toInt();
+}
+
+template<typename T> T value(const QVariant& variant) {
+    return variant.value<T>();
+}
```
patrickelectric: Should such functions be inside a namespace to avoid conflicts with others `values` functions/templates in the future ?

tcanabrava: or a qVariantValueOf()

PR #630, opened 2019-07-31
waterfallplot: remove signal override

tcanabrava:
mouseDepthChanged is declared in waterfall.h

### DIFF
--- a/src/waterfall/waterfall.h
+++ b/src/waterfall/waterfall.h
@@ -142,7 +142,6 @@ public:
 signals:
     void antialiasingChanged();
 
-    void mouseDepthChanged();
     void mouseConfidenceChanged();
     // TODO: mouseMove should be renamed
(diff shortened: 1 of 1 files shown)


src/waterfall/waterfallplot.h
```diff
@@ -118,7 +118,6 @@ class WaterfallPlot : public Waterfall
     void minDepthToDrawChanged();
     void mouseColumnConfidenceChanged();
     void mouseColumnDepthChanged();
-    void mouseDepthChanged();
```
patrickelectric: You are right, but the problem is different, the function should not exist in `waterfall.h` since it's only a base class for "waterfall like" plots. Such plot can not provide depth values, like what we are having in Ping360. 
TLDR: the problem is in waterfall base class, the signal should be removed there.

PR #632, opened 2019-07-31
Warning emit inside constructor

tcanabrava:


tcanabrava:
Emits should not happen in a constructor, there's no object connected to it yet.

PR #634, opened 2019-07-31
DeviceManager: Enable detection of Ping360 over ethernet

### LINKED ISSUE #601: ping360: implement ip address discovery via udp broadcast

Williangalvani:
Screenshot from 2019-07-31 17-16-02

fix #601

### DIFF
--- a/src/devicemanager/devicemanager.cpp
+++ b/src/devicemanager/devicemanager.cpp
@@ -21,6 +21,7 @@ DeviceManager::DeviceManager() :
         qCDebug(DEVICEMANAGER) << "Available devices:" << links;
         updateAvailableConnections(links);
     });
+    connect(&_ping360EthernetFinder, &Ping360EthernetFinder::availableLinkFound, this, &DeviceManager::append);
     append({AbstractLinkNamespace::Ping1DSimulation, {}, "Ping1D Simulation", PingDeviceType::PING1D}, "Ping1D");
     append({AbstractLinkNamespace::Ping360Simulation, {}, "Ping360 Simulation", PingDeviceType::PING360}, "Ping360");
 }
@@ -60,11 +61,13 @@ void DeviceManager::startDetecting()
 {
     qCDebug(DEVICEMANAGER) << "Start protocol detector service.";
     _detectorThread.start();
+    _ping360EthernetFinder.start();
 }
 
 void DeviceManager::stopDetecting()
 {
     qCDebug(DEVICEMANAGER) << "Stop protocol detector service.";
+    _ping360EthernetFinder.stop();
     _detector->stop();
     _detectorThread.quit();
 }
--- a/src/devicemanager/devicemanager.h
+++ b/src/devicemanager/devicemanager.h
@@ -5,6 +5,7 @@
 #include <QThread>
 
 #include "abstractlinknamespace.h"
+#include "ping360ethernetfinder.h"
 #include "protocoldetector.h"
 #include "sensor.h"
 
@@ -175,6 +176,7 @@ private:
 
     QSharedPointer<Sensor> _primarySensor;
     ProtocolDetector* _detector;
+    Ping360EthernetFinder _ping360EthernetFinder;
     QThread _detectorThread;
 
     // Model variables
--- a/src/sensor/ping360ethernetfinder.cpp
+++ b/src/sensor/ping360ethernetfinder.cpp
@@ -1,0 +1,45 @@
+#include <QUdpSocket>
+
+#include "ping360ethernetfinder.h"
+
+Ping360EthernetFinder::Ping360EthernetFinder()
+{
+    connect(&_broadcastTimer, &QTimer::timeout, this, &Ping360EthernetFinder::doBroadcast);
+    connect(&_broadcastSocket, &QUdpSocket::readyRead, this, &Ping360EthernetFinder::processBroadcastResponses);
+}
+
+void Ping360EthernetFinder::start()
+{
+    _broadcastTimer.start(1000);
+}
+
+void Ping360EthernetFinder::doBroadcast()
+{
+    // We broadcast a "Discovery" message to 255.255.255.255 and port 30303 as
+    // described in the communications manual.
+    QByteArray datagram = "Discovery";
+    _broadcastSocket.writeDatagram(datagram, QHostAddress::Broadcast, 30303);
+}
+
+void Ping360EthernetFinder::processBroadcastResponses()
+{
+    while(_broadcastSocket.hasPendingDatagrams()) {
+        QHostAddress sender;
+        QByteArray datagram;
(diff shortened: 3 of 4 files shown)


Williangalvani:
There is a small issue. It keeps alternating between green and red:
ping360 intermitent

src/devicemanager/devicemanager.cpp
```diff
@@ -59,12 +61,16 @@ void DeviceManager::append(const LinkConfiguration& linkConf, const QString& dev
 void DeviceManager::startDetecting()
 {
     qCDebug(DEVICEMANAGER) << "Start protocol detector service.";
+    _broadcastTimer.start(1000);
+    _broadcastSocket.bind(QHostAddress::Any, 1048);
```
patrickelectric: Thy is this in device manager and no in protocoldetector ?
DeviceManager is just a class that manages the device creation from the links that are possible to connect, everything related to the discovery process should be in protocol detector

patrickelectric: Do we need to use `bind` ? From what we talked about bind is not necessary if the other side is a server and it'll probably be.

src/devicemanager/devicemanager.cpp
```diff
@@ -147,6 +153,28 @@ DeviceManager* DeviceManager::self()
     return self;
 }
 
+void DeviceManager::doBroadcast()
+{
+    QByteArray datagram = "Discovery";
```
patrickelectric: Why are you sending a "Discovery" string ?

src/devicemanager/devicemanager.cpp
```diff
@@ -147,6 +153,28 @@ DeviceManager* DeviceManager::self()
...
+    _broadcastSocket.writeDatagram(datagram, QHostAddress::Broadcast, 30303);
```
patrickelectric: Why we are using port 30303 ?

src/devicemanager/devicemanager.h
```diff
@@ -3,6 +3,8 @@
 #include <QAbstractListModel>
 #include <QLoggingCategory>
 #include <QThread>
+#include <QUdpSocket>
+#include <QTimer>
```
patrickelectric: Alphabetic order.

src/devicemanager/devicemanager.cpp
Williangalvani: My reasoning was that we don't need to detect the protocol. We know it is a Ping360. I can move it over, but I vote for renaming ProtocolDetector to DeviceDetector or something like that =]

Williangalvani: We should need it, we are doing a broadcast, there is no connection involved. We say "Yo, any Ping360 out there?" then they all reply to the message.
That said, I just tested and it worked, so I'm taking it out ¯\_(ツ)_/¯

src/devicemanager/devicemanager.cpp
Williangalvani: This is as documented in the "Communications Protocol v1.7.pdf"

src/devicemanager/devicemanager.cpp
Williangalvani: This is as documented in the "Communications Protocol v1.7.pdf"

src/devicemanager/devicemanager.cpp
```diff
@@ -147,6 +152,42 @@ DeviceManager* DeviceManager::self()
...
+}
+
+void DeviceManager::processBroadcastResponses()
+{
+    /*
+        We broadcast a "Discovery" message to 255.255.255.255 and port 30303 as
+        described in the communications manual.
+        The devices replies with IP and MAC addres in this format:
```
patrickelectric: Can something like this comment be add in the function description ?

Williangalvani:
You are right, it is not clear that it is for Ping360 only. I'll get it into it's own class tomorrow. Maybe Ping360Udp<Detector|Scanner|Finder> (hopefully I come up with a better name during my sleep).

src/devicemanager/devicemanager.cpp
patrickelectric: Indeed, We should rework and make it more abstract and clear what is the paper of each class.
Also, we should think more about the protocol detection side of things, allowing multiple protocols, different proprietary sensors and etc.

Williangalvani:
@patrickelectric I moved it to a class

src/devicemanager/devicemanager.h
```diff
@@ -175,6 +176,7 @@ class DeviceManager : public QAbstractListModel
 
     QSharedPointer<Sensor> _primarySensor;
     ProtocolDetector* _detector;
+    Ping360EthernetFinder* _ping360EthernetFinder;
```
patrickelectric: Is there a need for this to be a pointer ?

src/sensor/ping360ethernetfinder.cpp
```diff
@@ -0,0 +1,45 @@
...
+    while(_broadcastSocket.hasPendingDatagrams()) {
+        QHostAddress sender;
+        QByteArray datagram;
+        datagram.resize(_broadcastSocket.pendingDatagramSize());
+        _broadcastSocket.readDatagram(datagram.data(), datagram.size(), &sender);
+        // Make sure we have an IPV4 address, and not something like "::ffff:192.168.1.1"
+        sender = QHostAddress(sender.toIPv4Address());
+        // Basic validation to check we didn't get some lost packet from wherever
+        if(datagram.contains("PING360")) {
+            emit availableLinkFound({LinkType::Udp, {sender.toString(), "12345"}, "BlueRov2 Ping360 Port", PingDeviceType::PING360},
```
patrickelectric: This is not BlueRov2 Ping360 port, the description should be fixed.

src/devicemanager/devicemanager.cpp
```diff
@@ -21,6 +22,9 @@ DeviceManager::DeviceManager() :
         qCDebug(DEVICEMANAGER) << "Available devices:" << links;
         updateAvailableConnections(links);
     });
+    connect(_ping360EthernetFinder, &Ping360EthernetFinder::availableLinkFound, this, [this](auto link, auto name) {
+        append(link, name);
```
patrickelectric: do you need this lambda ? It appears that you can connect with append directly.

src/devicemanager/devicemanager.h
Williangalvani: nope, fixed.

src/devicemanager/devicemanager.h
```diff
@@ -6,6 +6,7 @@
 
 #include "abstractlinknamespace.h"
 #include "protocoldetector.h"
+#include "ping360ethernetfinder.h"
```
patrickelectric: alphabetic order

Williangalvani: done

PR #639, opened 2019-08-01
pingslider: add an editable spinbox if valueText is null

tcanabrava:


patrickelectric:
I have mixed fillings about it, it does not follow the same UI grid and the controls are pink (?). Appears to be weird IMO.
image

tcanabrava:
The colors are wrong, but it enables desktop and mobile users to click /
fine tune and input text.

Em sex, 2 de ago de 2019 às 14:27, Patrick José Pereira <

jaxxzer:
The function is good!

One thing that will help: highlight all text when clicked (or maybe double-clicked) for text entry.

Maybe it will look better if the -/+ icons are next to each other on the right hand side of the text entry.

jaxxzer:
> Range should receive users inputs, this should be predefined like we have already in master.

I think it is ok for the user to enter any number they want for range in the new manual entry fields, the default value snapping on the slider just make it easy for normal situations.

jaxxzer:
Whenever the range is adjusted, the transmit_duration should be adjusted automatically. The user may also change the transmit_duration to a specific value after they have selected the range.

There were some problems with this, my commit on top attempts to fix this. You can see in this example there is still a problem. Everything works correctly until I click the SpinBox. After I click the SpinBox, the automatic adjustment with range stops working:

Peek 2019-08-05 17-23

jaxxzer:
Whenever the range is adjusted, the transmit_duration should be adjusted automatically. The user may also change the transmit_duration to a specific value after they have selected the range.
There were some problems with this, my commit on top attempts to fix this. You can see in this example there is still a problem. Everything works correctly until I click the SpinBox. After I click the SpinBox, the automatic adjustment with range stops working:

 Peek 2019-08-05 17-23

tcanabrava:
On it.

Em seg, 5 de ago de 2019 às 23:30, Jacob Walser <notifications@github.com>
escreveu:

qml/PingSlider.qml
```diff
@@ -27,8 +27,21 @@ RowLayout {
...
+        id: spinBox
+        Layout.minimumWidth: 150
+        Layout.maximumWidth: 150
+        visible: valueText === null
+        from: sliderControl.from
+        to: sliderControl.to
+        value: sliderControl.value
+        onValueChanged: sliderControl.value = value
+        stepSize: sliderControl.stepSize
+        editable: true
```
patrickelectric: From the PR title, how do you know that the input is numeric ? I don't see any check about that.
Should the title be "add an editable spinbox if valueText is null" ?

qml/Ping360ControlPanel.qml
```diff
@@ -88,39 +88,39 @@ GroupBox {
                 value: ping.transmit_duration
                 control.from: ping.transmitDurationMin
                 control.to: ping.transmitDurationMax
-                control.onMoved: ping.transmit_duration = control.value
+                control.onValueChanged: ping.transmit_duration = control.value
```
patrickelectric: why this one uses control and the others not ?

qml/PingSlider.qml
```diff
@@ -7,9 +7,30 @@ RowLayout {
...
     property alias control: sliderControl
-    property var valueText : null
+    property var valueText: null
+
+    onValueChanged: {
+        console.log("-----")
+        console.log("Current values: min" + from + " max" + to + "value " + value)
+        console.log("SpinBox From:" + spinBox.from + " to " + spinBox.to + " value: " + spinBox.value + " currentValue " + value)
+        console.log("sliderControl From:" + sliderControl.from + " to " + sliderControl.to + " value: " + sliderControl.value + " currentValue " + value)
+        console.log("-----")
```
patrickelectric: Do not commit debug code.

qml/PingSlider.qml
```diff
@@ -7,9 +7,30 @@ RowLayout {
...
+
+        if (spinBox.to < value) {
```
patrickelectric: should we do the same check for `from` ?

qml/PingSlider.qml
```diff
@@ -49,8 +48,27 @@ RowLayout {
...
+
+    Binding on value {
+        when: spinBox.down.pressed
+        value: spinBox.value
+    }
+
+    Binding on value {
+        when: spinBox.focus
+        value: spinBox.value
+    }
```
patrickelectric: I believe that all spinbox binds could be replaced by `valueModified` signal

qml/PingSlider.qml
```diff
@@ -49,8 +48,27 @@ RowLayout {
...
         to: root.to
-        onValueChanged: root.value = value
         stepSize: sliderControl.stepSize
         editable: true
     }
+
+    Binding on value {
+        when: sliderControl.pressed
+        value: sliderControl.value
+    }
```
patrickelectric: Since this is control specific, should this be inside sliderControl element ?

qml/PingSlider.qml
tcanabrava: Okay.

qml/PingSlider.qml
tcanabrava: this is control specific, it's taking care of the root control, changing the root.value when sliderControl is pressed.

qml/PingSlider.qml
tcanabrava: I tried. does not work. the problem is that:

onValueModified: root.value = value <--- this breaks the binding

creating a Binding on value {} doesn't breaks dynamic bindings.

patrickelectric: I was talking about:
```
Binding on value {
	when: spinBox.valueChanged
    ...
}
```

PR #640, opened 2019-08-01
devicemanagerviewer: remove the 'Connect' tooltip

tcanabrava:
The connect tooltip behaved strangely: it appeared in a unrelated
position of the list and had a "ghost" that took one to two seconds
to disappear after the window disappears.

For that reason, I removed the tooltip and changed the background
color of the device to hiligth that this is clickable


patrickelectric:
It's not clear that the available devices is a list with this visual changes, we need something to help with that like some horizontal lines between each component or in the top and bottom of the listview

image

tcanabrava:
close for splitting into more pr`s.

Issue #603, opened 2019-07-26
Improve PingStatus debug windows


jaxxzer:
It's a bit annoying for the sensor debug window to be fixed in the center of the visualizer. I think we should move it to bottom-left and/or allow dragging. It will also be cool if we can spice the style with some cool monospace programmer font in lime green.

![Screenshot_20190726_113246](https://user-images.githubusercontent.com/8315108/61963216-19d8b380-af99-11e9-96e3-bb8be059eb5e.png)

patrickelectric:
Fixed in #651

PR #643, opened 2019-08-02
Move debug overlays to lower left corner

Williangalvani:
The debug window on top of the plot is really annoying. this moves it to the lower left corner:
Screenshot from 2019-08-02 14-50-14

Screenshot from 2019-08-02 14-54-16

### DIFF
--- a/qml/Ping360Status.qml
+++ b/qml/Ping360Status.qml
@@ -10,7 +10,8 @@ Item {
     property var ping: null
 
     Rectangle {
-        anchors.centerIn: parent
+        anchors.left: parent.left
+        anchors.bottom: parent.bottom
 
         color: "black"
         opacity: 0.75
--- a/qml/PingStatus.qml
+++ b/qml/PingStatus.qml
@@ -10,7 +10,8 @@ Item {
     property var ping: null
 
     Rectangle {
-        anchors.centerIn: parent
+        anchors.left: parent.left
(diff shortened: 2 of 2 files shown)


patrickelectric:
Right now the debug widget is over the replay menu button.
image

jaxxzer:
> problem with replay button

what do you suggest? Is it really a problem?

Williangalvani:
I think it would be nice to have it draggable around the screen, but I'm not sure of how to do that with qml

patrickelectric:
> Is it really a problem? What do you suggest to resolve it?

Yes, since it is over the control button.

> I think it would be nice to have it draggable around the screen, but I'm not sure of how to do that with qml

Check `ValueReadout.qml`.

jaxxzer:
> Yes, since it is over the control button.

I think that the patch is doing better than what we have today. Right now the item is over the main plot that you want to see. The collision of items in this patch is much less obstructive than we have in master, and this patch fixes #603 so why can't we merge it now? You can put another issue or push your own commits to improve it further. Right now we have a patch to improve master, ready to merge.

jaxxzer:
> Right now we have a patch to improve master, ready to merge.

And we should generally merge things like that instead of waiting one week or more for something better to happen.

patrickelectric:
Done in #651

PR #644, opened 2019-08-05
devicemanagerviewer: add a single line between delegates

tcanabrava:
Screenshot_20190805_171257

### DIFF
--- a/qml/DeviceManagerViewer.qml
+++ b/qml/DeviceManagerViewer.qml
@@ -130,15 +130,25 @@ PingPopup {
                 Layout.fillHeight: true
                 clip: true
                 model: DeviceManager
-
+                spacing: 4
                 delegate: Rectangle {
+                    id: rectDelegate
                     height: 40
                     width: parent.width
-                    border.width: 1
-                    border.color: "gray"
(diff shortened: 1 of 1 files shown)


tcanabrava:
> It does not separate each element.
> image

yeah, I had problems using DeviceManager.count - the property for it was emiting without a value. :)

qml/DeviceManagerViewer.qml
```diff
@@ -154,7 +164,6 @@ PingPopup {
                             text: "Connect."
                         }
                     }
-
```
patrickelectric: Avoid style changes in functionality changes.

src/devicemanager/devicemanager.cpp
```diff
@@ -54,7 +54,7 @@ void DeviceManager::append(const LinkConfiguration& linkConf, const QString& dev
     const auto& indexRow = index(line);
     endInsertRows();
     emit dataChanged(indexRow, indexRow, _roles);
-    emit countChanged();
+    emit countChanged(_sensors[Available].size());
```
patrickelectric: this should and it's not necessary.

PR #645, opened 2019-08-05
pingproperties: move the properties macro elsewhere

tcanabrava:
Now we can start using it thru the code

### DIFF
--- a/src/util/pingproperties.h
+++ b/src/util/pingproperties.h
@@ -1,0 +1,27 @@
+#pragma once
+
+#include <QtGlobal>
+
+/** Calls the correct operator == in the P_PROPERTY macro */
+template<typename T> bool pingIsSame(const T& a, const T& b)
+{
+    return a == b;
+}
+
+/** Calls the correct operator == in the P_PROPERTY macro */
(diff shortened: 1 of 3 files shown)


src/util/pingproperties.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include <QtGlobal>
+
+/** Calls the correct operator == in the P_PROPERTY macro */
+template<typename T> bool pingIsSame(const T& a, const T& b)
```
patrickelectric: how is this different from the cpp ones ?
Like, why this definition is here and the others in the cpp ?

tcanabrava: because this is a generic template. What happens is when you have a non specialized template, the code can't be in the .cpp as the code will be generated during compilation. 

You can have three possibilities with templates:
- No specializations at all: all of the code must be in the header
- Partial specialization: if at least *one* of the template parameters is a template generic type, all the code must be in the header.
- Full specialization: if the code is fully specialized (no template parameter at all), the code cannot be in the header as multiple definition errors would occour in the link time.


The first one defines a template, the other two fully specializes it.

PR #651, opened 2019-08-06
Drag drop status

tcanabrava:
Make the Status panels movable in the ping1d and ping 360

### DIFF
--- a/qml/Ping360Status.qml
+++ b/qml/Ping360Status.qml
@@ -10,12 +10,23 @@ Item {
     property var ping: null
 
     Rectangle {
-        anchors.centerIn: parent
-
+        id: rect
         color: "black"
         opacity: 0.75
         height: innerCol.height + 2 * marginPix
         width: innerCol.width + 5 * marginPix
+        x: root.width / 2
+        y: root.height / 2
+
+        MouseArea {
+            anchors.fill: parent
+            drag.target: rect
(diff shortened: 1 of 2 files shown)


qml/Ping360Status.qml
```diff
@@ -10,12 +10,20 @@ Item {
...
+        id: rect
         color: "black"
         opacity: 0.75
         height: innerCol.height + 2 * marginPix
         width: innerCol.width + 5 * marginPix
+
+        MouseArea {
+            anchors.fill: parent
+            drag.target: rect
+            drag.minimumX: 0
```
patrickelectric: Its missing `minimumY`

qml/Ping360Status.qml
```diff
@@ -10,12 +10,20 @@ Item {
     property var ping: null
 
     Rectangle {
-        anchors.centerIn: parent
-
+        id: rect
```
patrickelectric: It would be nice to have x and y equal to `x: (root.width - rect.width)/2 and y: 0`, right now the rect starts in the top left corner.

patrickelectric: image

tcanabrava: I'll do. I tried to do something similar but the root triggered a change in width during the drag while the drawing was updating and that made the rectangle jump.

PR #638, opened 2019-08-01
Combo slider ticks

tcanabrava:
Create Ticks and Labels for the PingSlider
The text positioning is a bit off, I appreciate some hints on the math.

PR #662, opened 2019-08-08
Corrections around #638

patrickelectric:
- Use color property
- Remove hardcoded values
- Style corrections
- Remove unused variables


patrickelectric:
It was not only the commits organization that I have done some work, others corrections was also done. Just some small stuff to avoid extra work/reviews.

PR #666, opened 2019-08-09
Improved speed: Updates over #540

patrickelectric:
- Small improvements over #540 
- Corrections/Quality of life in ABR

I can break this in two PRs, but it's easier to test the speed improvements with the ABR improvements. 

This PR should hit something near 60~50Hz with the timer latency configured.
Linux: `sudo echo 1 > /sys/bus/usb-serial/devices/ttyUSB0/latency_timer`
Windows: **Device Manager** → Com port → **Properties** → **Advanced** → Latency Timer: 1

It takes 8s  to do a full spin in a 2m range and 2M baud rate.
> Note: Message frequency is not the same thing as full scan time. Messages in the same position will be faster and with more messages but with slow scan time. We should add scan time in the sensor status.

### DIFF
--- a/src/sensor/ping360.cpp
+++ b/src/sensor/ping360.cpp
@@ -320,7 +320,6 @@ void Ping360::setBaudRate(int baudRate)
 
     qCDebug(PING_PROTOCOL_PING360) << "Moving to baud rate:" << baudRate;
     serialLink->setBaudRate(baudRate);
-    link()->startConnection();
     emit linkUpdate();
 }
 
--- a/src/sensor/ping360.cpp
+++ b/src/sensor/ping360.cpp
@@ -163,13 +163,12 @@ void Ping360::handleMessage(const ping_message& msg)
         if(_configuring) {
             _baudrateConfigurationTimer.start();
             checkBaudrateProcess();
-            return;
         } else {
             _baudrateConfigurationTimer.stop();
             _timeoutProfileMessage.start();
             requestNextProfile();
-            return;
         }
+        return;
     }
 
     case Ping360Id::DEVICE_DATA: {
--- a/src/sensor/ping360.cpp
+++ b/src/sensor/ping360.cpp
@@ -151,12 +151,6 @@ void Ping360::handleMessage(const ping_message& msg)
 {
     qCDebug(PING_PROTOCOL_PING360) << "Handling Message:" << msg.message_id();
 
-    // Update frequency for each
-    messageFrequencies[msg.message_id()].setElapsed(_messageElapsedTimer.elapsed());
-    // Since we don't have a huge number of messages and this variable is pretty simple,
-    // we can use a single signal to update someone about the frequency update
-    emit messageFrequencyChanged();
-
     switch (msg.message_id()) {
 
     case CommonId::DEVICE_INFORMATION: {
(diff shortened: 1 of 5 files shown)


Williangalvani:
Test results here:
**Serial**
 -  windows up to 50hz
 -   linux up to 60hz
**ethernet**
 - windows up to 80hz
 - linux TBA (rebooting)

src/sensor/ping360.cpp
```diff
@@ -80,16 +80,20 @@ void Ping360::startPreConfigurationProcess()
...
     static int count = _ABRTotalNumberOfMessages;
 
+    if(reset) {
+        count = _ABRTotalNumberOfMessages;
+    }
+
     // We are starting to check the new baud rate
     if(count == _ABRTotalNumberOfMessages) {
-        detectBaudrates();
+        detectBaudrates(reset);
```
jaxxzer: should this be?

```suggestion
        detectBaudrates(true);
```

patrickelectric: This will reset the function each time, making it back to the highest baudrate. The idea of reset is to restart all the procedure.
The reset is only true at the first time that the detection run or after an user request.

src/sensor/ping360.cpp
```diff
@@ -151,28 +155,28 @@ void Ping360::handleMessage(const ping_message& msg)
...
             requestNextProfile();
-            return;
         }
+        return;
     }
 
     case Ping360Id::DEVICE_DATA: {
+        // Request next message ASAP
+        // request another transmission
+        requestNextProfile();
```
jaxxzer: I thought we needed to extract the current device angle from the device_data message before calling this.

patrickelectric:
@jaxxzer 	right now everything runs in the same thread, the only link that contains a thread is the filelink, but there is no directConnection on it.
I'll add @tcanabrava 	to review it also.

src/sensor/ping360.cpp
patrickelectric: Done.

src/sensor/pingsensor.cpp
```diff
@@ -10,7 +10,8 @@ PingSensor::PingSensor()
     ,_lostMessages(0)
 {
     _parser = new PingParserExt();
-    connect(dynamic_cast<PingParserExt*>(_parser), &PingParserExt::newMessage, this, &PingSensor::handleMessagePrivate);
+    connect(dynamic_cast<PingParserExt*>(_parser), &PingParserExt::newMessage, this, &PingSensor::handleMessagePrivate,
```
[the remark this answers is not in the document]
patrickelectric: This PR is already complex in the way that it's. This update only improves the connection and message handle speed and not the `connection` call speed itself. Since this is done in the constructor and it only happens when the device is connected, this cast time is not a huge concern right now, we can deal with it in future PRs.

src/sensor/ping360.cpp
```diff
@@ -80,16 +80,20 @@ void Ping360::startPreConfigurationProcess()
     writeMessage(msg);
 }
 
-void Ping360::checkBaudrateProcess()
+void Ping360::checkBaudrateProcess(bool reset)
```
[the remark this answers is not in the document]
patrickelectric: Moved to class variable.

patrickelectric:
@tcanabrava 	changes done.

Williangalvani:
Latest version is reaching 60hz in Linux via UDP, 6.6s for full scan

Issue #665, opened 2019-08-09
Set serial driver latency timer programatically + automatically


jaxxzer:
- [x] Windows
 - [x] Linux
 - [ ] Mac

PR #673, opened 2019-08-13
Configure serial latency for the lowest possible

patrickelectric:
Helps #665 
For windows and linux

### DIFF
--- a/src/link/seriallink.cpp
+++ b/src/link/seriallink.cpp
@@ -6,6 +6,16 @@
 
 #include "seriallink.h"
 
+#ifdef Q_OS_LINUX
+    #include <sys/ioctl.h>
+    #include <termios.h>
+    #include <linux/serial.h>
+#endif
+#ifdef Q_OS_WIN
+    #include <Windows.h>
+    #include <QSettings>
+#endif
+
 Q_LOGGING_CATEGORY(PING_PROTOCOL_SERIALLINK, "ping.protocol.seriallink")
 
 SerialLink::SerialLink(QObject* parent)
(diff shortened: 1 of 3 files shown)


src/link/seriallink.cpp
```diff
@@ -6,6 +6,16 @@
 
 #include "seriallink.h"
 
+#ifdef Q_OS_LINUX
+    #include <sys/ioctl.h>
+    #include <termios.h>
+    #include <linux/serial.h>
+#endif
+#ifdef Q_OS_WIN
+    #include <Windows.h>
```
tcanabrava: windows.h - non capitals. Windows is moving torwards to a case sensitive file system in the future.

src/link/seriallink.cpp
```diff
@@ -99,12 +109,146 @@ void SerialLink::setBaudRate(int baudRate)
...
+    }
+    startConnection();
+#endif
+
+    //light at the end of the tunnel
+    return true;
+
+#endif
+}
+
```
tcanabrava: break that function into two, one that handles windows, another one that handles linux? it's hard to read with the ifdef blogs getting the whole function.

patrickelectric: I would like to avoid such thing, because this will need ifdefs in header and ifdefs in cpp. Let me know if you a better solution.

src/link/seriallink.cpp
patrickelectric: From QtCreator:
image

Should we use `Windows` or `windows` ?

Williangalvani:
I can't get it to work under Windows:
These are 3 attempts, the last one I right-clicked to give administrator privileges, but still no success
[20190814-132636806.txt](https://github.com/bluerobotics/ping-viewer/files/3502325/20190814-132636806.txt)
[20190814-132756703.txt](https://github.com/bluerobotics/ping-viewer/files/3502323/20190814-132756703.txt)
[20190814-132832858.txt](https://github.com/bluerobotics/ping-viewer/files/3502324/20190814-132832858.txt)

patrickelectric:
@Williangalvani 	it appears that you are compiling the wrong version. From your log you are building the commitid `9cf14dd`.
`[13:26:40:463] ping.main[Info]: Git version: 9cf14dd`

Williangalvani:
I'm actually getting the [appveyor](https://ci.appveyor.com/project/blueroboticsappveyor/ping-viewer/builds/26693101) artifact. I'm not sure why it shows a different hash... Any ideas?

jaxxzer:
> I'm actually getting the [appveyor](https://ci.appveyor.com/project/blueroboticsappveyor/ping-viewer/builds/26693101) artifact. I'm not sure why it shows a different hash... Any ideas?

appveyor creates a [temporary merge commit](https://github.com/bluerobotics/ping-viewer/commit/9cf14dde9b5844a821568f9ca6024cb91ee56b68).

Screenshot_20190814_142809

patrickelectric:
@Williangalvani 	can you test it now ?

Williangalvani:
Looks good now, full scan time went down from 13s to 10s.

src/link/seriallink.cpp
tcanabrava: then  I got it wrong and the case of the Windows.h is actually in capitals.

patrickelectric:
@tcanabrava 	@Williangalvani 	I broke the function in 3.

PR #677, opened 2019-08-15
Fix preconfiguration in ping360 for non writable or serial connections

patrickelectric:


src/sensor/ping360.cpp
```diff
@@ -73,6 +73,20 @@ Ping360::Ping360()
 
 void Ping360::startPreConfigurationProcess()
 {
+    // Stop all configuration/message timers if link is not writable
```
Williangalvani: is this for replaying logs? What kinds of links are not writeable?

patrickelectric: yes, log is also non writable.

patrickelectric:
@Williangalvani 	can you confirm if this PR is working for you ?

Williangalvani:
> @Williangalvani can you confirm if this PR is working for you ?

It worked fine for udp, serial, and replay in linux. I have to do some tech support now so I can't test Windows right away

src/sensor/ping360.cpp
```diff
@@ -73,6 +73,20 @@ Ping360::Ping360()
...
+    if(!checkForWitableLinkAndDisableConfiguration()) {
```
Williangalvani:
```suggestion
    if(!checkForNonWritableLinkAndDisableConfiguration()) {
```

patrickelectric: Done.

patrickelectric:
@Williangalvani 	corrections done.

src/sensor/ping360.cpp
Williangalvani: Ok, this doesn't look right, now. Did you change other stuff too?
Now the name does not agree with the returns! Sorry if this was my fault.

patrickelectric: I Changed the name again, for isWritable. This make it explicit what is happening.
Mostly of the cases we check for nonWritable, but the negative logic returned from a function makes it harder to understand since it's more implicit what is happening.

src/sensor/ping360.cpp
```diff
@@ -73,6 +73,20 @@ Ping360::Ping360()
 
 void Ping360::startPreConfigurationProcess()
 {
+    // Stop all configuration/message timers if link is not writable
+    if(!isWitableLinkAndDisableConfiguration()) {
```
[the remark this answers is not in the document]
patrickelectric: done.

PR #680, opened 2019-08-17
Cmake port

tcanabrava:
Initial port of ping-viewer to CMake, it builds, and we can see a few problems with the dependency graph of the software by looking at the libraries that each internal library links against. I also added a few forced defines to CMake to block some misuse of Qt, and because of that I had to work thru the headers and code.

### DIFF
--- a/src/filemanager/filemanager.cpp
+++ b/src/filemanager/filemanager.cpp
@@ -3,7 +3,7 @@
 #include <QStandardPaths>
 
 #include "filemanager.h"
-#include "logger.h"
+#include "logger/logger.h"
 
 PING_LOGGING_CATEGORY(FILEMANAGER, "ping.filemanager");
 
--- a/src/filemanager/filemanager.cpp
+++ b/src/filemanager/filemanager.cpp
@@ -24,7 +24,7 @@ FileManager::FileManager()
         if(!f->dir.exists()) {
             qCDebug(FILEMANAGER) << "Create folder" << f->dir.path();
             f->ok = rootDir.mkpath(f->dir.path());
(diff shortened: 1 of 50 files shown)


src/filemanager/filemanager.cpp
```diff
@@ -24,7 +24,7 @@ FileManager::FileManager()
         if(!f->dir.exists()) {
             qCDebug(FILEMANAGER) << "Create folder" << f->dir.path();
             f->ok = rootDir.mkpath(f->dir.path());
-            qCDebug(FILEMANAGER) << (f->ok ? "Done." : ("Error while creating folder" + f->dir.path()));
+            qCDebug(FILEMANAGER) << (f->ok ? QStringLiteral("Done.") : ("Error while creating folder" + f->dir.path()));
```
patrickelectric: Is there any reason to use QStringLiteral here and not in the else ?

patrickelectric:
I believe that three things are necessary to merge this patch:
	1. Pass the CI
	2. Add cmake compilation in travis as a new job 
	3. New issues in the repository to list what is necessary to finish the cmake integration.

Things that are missing but can be done in future PRs:
	1. It's missing 'requireAdministrator' for windows
	2. The icons for windows and mac release
	3. Needs git definitions for the build 
	4. Missing compile messages to help debug compile problems
    5. Add appveyor build
	6. Add Linux, macOS and Windows deployment

tcanabrava:
No problem, I’ll try to add them.


Em seg, 19 de ago de 2019 às 21:25, Patrick José Pereira <

PR #689, opened 2019-08-21
pingslider: dont enable controls when not visible

tcanabrava:
Theres a bud in qml that's activating some controls even when they
are not visible in the scene


patrickelectric:
@tcanabrava I'm unable to replicate the issue that this PR is trying to fix, can you say how are you able to replicate it under master ?

PR #704, opened 2019-08-22
Add user feedback when screenshot is taken 

### LINKED ISSUE #510: Create user feedback when hotkey is used

patrickelectric:
Fix #510


Williangalvani:
This is pausing the scanning very briefly (1 second?) here. I like that it stops for a little because a print is a still, but the scanning itself stopping is a little weird.

patrickelectric:
@Williangalvani I was not able to replicate it here with mac and linux.

Williangalvani:
> @Williangalvani I was not able to replicate it here with mac and linux.

I just realized this was present before this pr, merge at will!
To replicate it, just press 'w' while looking at the sonar head.

PR #705, opened 2019-08-22
Ping360Visualizer: Makes vertical and horizontal flip mutually exclusive for 360 sector

### LINKED ISSUE #701: S10: Makes vertical and horizontal flip mutually exclusive for 360 viewer

patrickelectric:
Fix #701 

### DIFF
--- a/qml/Ping360Visualizer.qml
+++ b/qml/Ping360Visualizer.qml
@@ -209,23 +209,19 @@ Item {
     Component {
         id: displaySettings
         GridLayout {
+            id: settingsLayout
             anchors.fill: parent
             columns: 5
             rowSpacing: 5
             columnSpacing: 5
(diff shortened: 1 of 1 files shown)


Williangalvani:
The behavior is inconsistent:
Peek 2019-08-23 18-17

when you have the top one selected and click the bottom one, the top is unchecked and the bottom is checked.
when you have the bottom one checked, the top one is disabled.

patrickelectric:
@Williangalvani done, it's not possible to accomplish the same thing with buttongroup, since there is 3 things changing how vertical flip works. there is also the problem that when the button is not visible, the enabled component of such control is also false, so it makes the onEnabled logic harder. This approach uses the binding method to accomplish the same behavior;

PR #725, opened 2019-08-26
ping360: Improve pulse duration equation

### LINKED ISSUE #702: L1: pulse length should be improved
> "After some experimentation with the sonar in the Marina I think the algorithm
for setting the pulse length could do with multiplying by 2 so that the pulse
length in auto mode is 54usec rather than the current 27usec."

patrickelectric:
Duplicate pulse duration as requested to improve visualization
It would be nice if someone could test that in the water 

Fix #702 

### DIFF
--- a/src/sensor/ping360.h
+++ b/src/sensor/ping360.h
@@ -404,16 +404,16 @@ public:
         if (_autoTransmitDuration) {
             /*
              * Per firmware engineer:
-             * 1. Starting point is TxPulse in usec = ((one-way range in metres) * 4000) / (Velocity of sound in metres per second)
+             * 1. Starting point is TxPulse in usec = ((one-way range in metres) * 8000) / (Velocity of sound in metres per second)
              * 2. Then check that TxPulse is wide enough for currently selected sample interval in usec, i.e.,
-             *    if TxPulse < (1.25 * sample interval) then TxPulse = (1.25 * sample interval)
+             *    if TxPulse < (2.5 * sample interval) then TxPulse = (2.5 * sample interval)
              * 3. Perform limit checking
              */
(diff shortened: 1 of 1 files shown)


src/sensor/ping360.h
```diff
@@ -404,16 +404,16 @@ class Ping360 : public PingSensor
...
+             * 1. Starting point is TxPulse in usec = ((one-way range in metres) * 8000) / (Velocity of sound in metres per second)
              * 2. Then check that TxPulse is wide enough for currently selected sample interval in usec, i.e.,
-             *    if TxPulse < (1.25 * sample interval) then TxPulse = (1.25 * sample interval)
+             *    if TxPulse < (2.5 * sample interval) then TxPulse = (2.5 * sample interval)
              * 3. Perform limit checking
              */
 
             // 1
-            int autoDuration = round(4000*range()/_speed_of_sound);
+            int autoDuration = round(8000*range()/_speed_of_sound);
```
tcanabrava: const int autoDuration = std::max(static_cast<int>(2.5*samplePeriod()/1000), round(4000*range()/_speed_of_sound));

define / set once

patrickelectric: I think the logic behind auto duration is already complex in the way that it's. We don't have the knowledge behind this steps used in this code, and I would like to maintain it as simple as possible. 
Besides that, this PR only duplicates the values, it's not supposed to change the logic behind it.

jaxxzer:
> It would be nice if someone could test that in the water

attention @brianhBR @kklemens We are changing the automatic pulse duration on recommendation that it provides better results. Please let us know how it goes after this commit.

PR #736, opened 2019-08-26
Fix debug window to show log playback information 

### LINKED ISSUE #727: M4: debug panel values do not match log playback
Example file: the range is changed several times, but the debug panel does not reflect this:
[test.txt](https://github.com/bluerobotics/ping-viewer/files/3542247/test.txt)

patrickelectric:
Fix #727


patrickelectric:
@Williangalvani good catch, number_of_points added, but angle offset is not possible since it's not a sensor configuration but a visualization settings.

PR #752, opened 2019-09-19
PingStatus: Disable models if parent is not visible

### LINKED ISSUE #750: ping1d performance suffers on t2
t2.0.2 framerate is low and very jittery. The display does not update smoothly.

@50ping/s setting with no transducer plugged in, this is on a direct usb connection:
t2.0.2: 88% cpu
v1.1.1: 44% cpu

patrickelectric:
Avoid unnecessary CPU usage and realocation of each new signal for each property

Fix #750

### DIFF
--- a/qml/PingStatus.qml
+++ b/qml/PingStatus.qml
@@ -11,8 +11,8 @@ Item {
 
     property var marginPix: 10
     property var ping: DeviceManager.primarySensor
-
     property var sensorModel: ping.sensorStatusModel(root)
+
     DelegateModel {
         id: baseModel
         property var title: "Base data:"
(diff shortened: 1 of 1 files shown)


qml/PingStatus.qml
```diff
@@ -62,15 +62,17 @@ Item {
             Label {
                 text: baseModel.title
             }
+            //We disable the models with visibility to avoid unnecessary CPU usage in realocation
+            //TODO: We should improve it to remove the amount of CPU and realocation when visible
```
jaxxzer: what is the reallocation that is happening? it looks like the model is only updated once on L14. could we use assignment instead of binding?

jaxxzer: property var sensorModel = ping.sensorStatusModel(root)

patrickelectric: The problem is that the properties changes, each profile message results in 8 signals, in 50 Hz this results in 400 models constructions.  It appears that Qt delete and create a new model each time.

jaxxzer: > It appears that Qt delete and create a new model each time.

that is funny.

Issue #667, opened 2019-08-09
UI adjustments


jaxxzer:
The engineers at the lab have some requests on adjustments to the UI

@brianhBR please list them here.

brianhBR:
UI Comments:
=============
We recently went through the UI with the goal to see if there were any things that we thought might be better laid out. We are totally open to debating this and want to hear your thoughts. If anything is unclear I'd be happy to post a terribly drawn sketch. 

Also, I think it would be cool to add more tool tips to menu icons and sonar settings. I think the overall goal is that this software be as close to not needing a manual as possible.

Controls Page (Top menu with Arrow):
-----------------------------------
- Replace the wrench with an arrow similar to the one used to open the menu. It should point down with the advanced controls menu closed and up with it open
- Move the firmware update icon to it's own sub-menu on the gear settings menu. Also, maybe we replace the icon with something more like an update icon? ![download](https://user-images.githubusercontent.com/28631375/63385115-2a403c00-c355-11e9-8947-2bd4acb4231f.png)

- Add a dropdown to set mount orrientation (either head up or head down). This is to replace flip vertically or flip horizontally on the display page. 
- Move Auto Baud and baud setting to Device manager
- Add "reset to defaults" button
- Drop Angle offset - I don't think anyone will need this and it is a little confusing
- Try centering sliders and increase the padding on the menu? It looks a little off balance. When Receiver gain is set to medium it almost touches the border.

Settings Page (Gear): 
---------------------
- Change order to Device Manager, Display settings, Firmware update
- Add an option to hide the A-Scan on the display settings. Its very cool to have the A-Scan but in situations with limited screen area the circular plot is the priority.

Device manager:
------------------
General:
- 3 visually similar pages: Auto, Manual, and Demo. They should each have the same layout with the BR logo at the top and some feedback about the connection status under the logo
- I'd propose that we drop the scrolling banner (talking with dolphins, etc.) and use moving elipses (. , .., ...) after waiting for devices. The scrolling banner is fun but might be frustrating if the user is having connection issues.
- Add the device manager logo to the top left corner of the window (helps to teach user that that logo gets you to the device manager)

Auto page:
- We should move the simulations to the demo page so they don't distract from actual connected devices
- The green and red lights are a cool but might be better. I've had an active connection while it shows red and no connection with it showing green. I think they should be rock solid reliable or maybe we drop it?

Manual page:
- Visually similar to the auto page
[...]

PR #760, opened 2019-09-20
Move from wrench to arrow for configuration menu

patrickelectric:
Helps #667

### DIFF
--- a/qml/Ping1DControlPanel.qml
+++ b/qml/Ping1DControlPanel.qml
@@ -83,19 +83,35 @@ ColumnLayout {
                     indicator.x: width - indicator.width - rightPadding
                 }
 
-
                 PingImage {
                     id: advancedSettingsButton
-                    source: StyleManager.configureIcon()
+                    source: StyleManager.arrowIcon()
                     height: 50
                     width: 50
                     selected: SettingsManager.enableSensorAdvancedConfiguration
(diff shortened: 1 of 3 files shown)


Williangalvani:
Could you make it only one color(blue) and make the background highlight on hover so we have more consistent buttons?
The slider buttons look good and are clearly buttons, the arrow not so much, and it looks out of place being black.
Peek 2019-09-23 11-37

patrickelectric:
All our ping buttons have highlights when clicked, you can check in the menu. We don't have hover functionality to change the background as the native buttons, that's probably what you are missing. 
Also, this uses the same concept with the already existent buttons, it does not change it just moves from wrench to an arrow as the engineer team asked for.
We can add mouse area for all buttons in a different PR.

patrickelectric:
@Williangalvani check: https://github.com/bluerobotics/ping-viewer/issues/767

patrickelectric:
@Williangalvani My ping-viewer does not have this hover background feature.

patrickelectric:
Peek 2019-09-23 12-13

jaxxzer:
>  make it only one color(blue)

@Williangalvani video shows blue / black arrow between the two states. Check that before merege. The background thing should not block like @patrickelectric  pointed.

PR #770, opened 2019-09-23
flasher: Search for stm32flash in OS before local path

### LINKED ISSUE #769: Look for stm32flash in system and not only in local path

patrickelectric:
Fix #769 

### DIFF
--- a/src/flash/flasher.cpp
+++ b/src/flash/flasher.cpp
@@ -7,6 +7,7 @@
 #include <QProcess>
 #include <QRegularExpression>
 #include <QSerialPortInfo>
+#include <QStandardPaths>
 #include <QThread>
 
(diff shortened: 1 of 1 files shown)


src/flash/flasher.cpp
```diff
@@ -64,17 +65,28 @@ void Flasher::setVerify(bool verify)
 
 QString Flasher::stm32flashPath()
 {
-#ifdef Q_OS_OSX
-    // macdeployqt file do not put stm32flash binary in the same folder of pingviewer
-    static QString absoluteBinPath = QCoreApplication::applicationDirPath() + "/../..";
+    static const QString binaryName =
```
jaxxzer: I think we should check local path first because it is already there for deployments and so that deployments are using a known version of stm32flash.

patrickelectric: Done.

PR #771, opened 2019-09-23
Fix linuxdeployqt

patrickelectric:
compile: Add libc6 copyright for linuxdeployqt

### DIFF
--- a/tools/compile.sh
+++ b/tools/compile.sh
@@ -197,10 +197,13 @@ echo ""
 echob "Start to deploy project.."
 
 if [[ $machine = *"Linux"* ]]; then
+    # For all crazy things necessary for linuxdeployqt, check https://github.com/probonopd/linuxdeployqt/issues/340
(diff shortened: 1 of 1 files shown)


jaxxzer:
Please add some comment about what/why and online reference if you have one.

patrickelectric:
https://github.com/probonopd/linuxdeployqt/issues/340

jaxxzer:
Wow long thread. For these kinds of problems when I had to go searching a long time for an answer, I like to add the link with reasoning + information directly inline with the text:

https://github.com/bluerobotics/battery-protector-firmware/blob/master/CMakeLists.txt#L11

Can you add some comment in the .sh.

probonopd:
> Wow long thread.

Boils down to: What goes into an AppImage should be compiled on a system no newer than the oldest still-supported version of Ubuntu, which is currently Ubuntu xenial.

> Add libc6 copyright for linuxdeployqt

is just satisfying the test for now, but that is not going to work much longer as  tests may be added. linuxdeployqt `-unsupported-...` flags shall not be used and may break at any time.

PR #778, opened 2019-09-25
Improve scan time

patrickelectric:
It reduces 360º@2m scan speed from 7.7~7.8s down to 7s.

### DIFF
--- a/pingviewer.pro
+++ b/pingviewer.pro
@@ -29,6 +29,7 @@ RESOURCES += \
 *msvc {
     QMAKE_CXXFLAGS += /WX \
         /wd4305 \ # Remove truncated warnings, msvc does not provide some non double values
+        /Ox \ # Enable speed optimizations
 
     # Necessary for register changes in serial layer for timer latency
     QMAKE_LFLAGS += /MANIFESTUAC:"level='requireAdministrator'"
--- a/pingviewer.pro
+++ b/pingviewer.pro
@@ -24,7 +24,7 @@ RESOURCES += \
 
 # Warning as error
 *-g++ | *-clang {
-    QMAKE_CXXFLAGS += -Werror
+    QMAKE_CXXFLAGS += -Werror -Ofast
 }
 *msvc {
     QMAKE_CXXFLAGS += /WX \
--- a/src/sensor/ping360.h
+++ b/src/sensor/ping360.h
@@ -63,18 +63,17 @@ public:
 
         int nextPoint = (_angle+delta)%_angularResolutionGrad;
 
-        ping360_transducer msg;
-        msg.set_mode(1);
-        msg.set_gain_setting(_gain_setting);
-        msg.set_angle(nextPoint);
-        msg.set_transmit_duration(_transmit_duration);
-        msg.set_sample_period(_sample_period);
-        msg.set_transmit_frequency(_transmit_frequency);
-        msg.set_number_of_samples(_num_points);
-        msg.set_transmit(transmit);
+        transducer_message.set_mode(1);
+        transducer_message.set_gain_setting(_gain_setting);
+        transducer_message.set_angle(nextPoint);
+        transducer_message.set_transmit_duration(_transmit_duration);
+        transducer_message.set_sample_period(_sample_period);
+        transducer_message.set_transmit_frequency(_transmit_frequency);
+        transducer_message.set_number_of_samples(_num_points);
+        transducer_message.set_transmit(transmit);
 
-        msg.updateChecksum();
-        writeMessage(msg);
+        transducer_message.updateChecksum();
+        writeMessage(transducer_message);
     }
 
     /**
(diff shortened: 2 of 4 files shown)


pingviewer.pro
```diff
@@ -24,11 +24,12 @@ RESOURCES += \
 
 # Warning as error
 *-g++ | *-clang {
-    QMAKE_CXXFLAGS += -Werror
+    QMAKE_CXXFLAGS += -Werror -Os
```
jaxxzer: this is optimization flag for code size

> It also enables -finline-functions, causes the compiler to tune for code size rather than execution speed

https://gcc.gnu.org/onlinedocs/gcc-9.2.0/gcc/Optimize-Options.html#Optimize-Options

src/sensor/pingsensor.cpp
```diff
@@ -39,7 +39,9 @@ void PingSensor::writeMessage(const ping_message& msg) const
 
 void PingSensor::handleMessagePrivate(const ping_message& msg)
 {
+#ifndef QT_NO_DEBUG
```
jaxxzer: I don't think we should repurpose this flag, it will loose asserts and other debugging things.

Try using QT_NO_DEBUG_OUTPUT, then we do not have to guard all of our qdebug statements.

Or make a new flag PING360_OPTIMIZE to guard here.

pingviewer.pro
patrickelectric: It appears that Ofast was lost somewhere in the rebase, fixed.

src/sensor/pingsensor.cpp
patrickelectric: We are just avoiding verbose messages in release that could impact performance.

src/sensor/ping360.h
```diff
@@ -63,7 +63,7 @@ class Ping360 : public PingSensor
 
         int nextPoint = (_angle+delta)%_angularResolutionGrad;
 
-        ping360_transducer msg;
+        static ping360_transducer msg;
```
jaxxzer: This is not good design if we ever instantiate multiple sensor classes. https://stackoverflow.com/a/6223371

You should move the 'ping360_transducer msg' to be a Ping360 class member variable, and then you can also call set_mode(1) only once in the constructor (one more little optimize).

src/sensor/pingsensor.cpp
jaxxzer: > We are just avoiding verbose messages in release that could impact performance.

That is what you mean to do, but you are also disabling any asserts() and so on that appear elsewhere in the program. The QT_NO_DEBUG flag is not designed for this purpose. You should have a PING360_OPTIMIZE flag that is independent of QT_NO_DEBUG.

jaxxzer: I see you are trying to target release builds only with this, where I guess the QT_NO_DEBUG flag is already set.

These debug messages are to help us find problems with users. We don't have debug build for mac and linux though. Also, I don't think it is a good trade to disable this for release builds.

patrickelectric: QT_NO_DEBUG is the "release" definition, there is no `QT_RELEASE` if that is what you are thinking. 


> The QT_NO_DEBUG flag is not designed for this purpose.
Can you point where is the purpose of such flag ? AFIK this flag is the "no" version of QT_DEBUG that is used to detect code compiled in debug mode. 

@tcanabrava any idea ?

jaxxzer: I see that this only covers release builds now @patrickelectric that is ok with me to use the same flag for that. Thank you for explaining it.

BUT I don't want to take these messages out of release builds. I approve this patch except for this commit, it should be removed.

patrickelectric: This messages are a bit annoying and also time consuming, removing this commit will result in almost 4% performance time, making the execution time from 7.7s down to 7.3s, with the others commits this brings down to 7s and with some that are still in development under 7s in execution time. This messages are not removed, if debug is necessary, you can use the debug builds, that's why we have it.

src/sensor/ping360.h
patrickelectric: Done.

src/sensor/pingsensor.cpp
tcanabrava: The issue is that even with the debug output removed with Q_NO_DEBUG_OUTPUT, the calculations done within it would still be done. I don't really like the #ifdef as it's hard to track the usecase when people compile it's own code, but we could do something in the lines of

if (debugCategoryIsEnabled) {
   qDebug() 
}

then it would be only a jmp and quite fast.

patrickelectric: With that such message will not saved in the log system.

jaxxzer: We cannot remove these messages at this point in time. I won't accept it. Perhaps later when everything is stable.

>  4% performance time lost, making the execution time from 7.7s down to 7.3s

This is not a good trade in my opinion. There are better changes we can make to improve scan speed drastically, without sacrificing our valuable logging debug information.

> you can use the debug builds, that's why we have it.

Our users cannot use debug builds, because we do not have it for mac and linux.

patrickelectric: We can do as Tomaz said, the messages will be there.

patrickelectric: I'll remove this commit for now to help this PR before next text release

PR #780, opened 2019-09-26
Find ping360 on secondary network

jaxxzer:
Broadcasting to 255.255.255.255 finds only responses from the default network interface (usually your router). With this patch we find it on the default interface + 192.168.2.x in case there are two network interfaces.

### DIFF
--- a/src/sensor/ping360ethernetfinder.cpp
+++ b/src/sensor/ping360ethernetfinder.cpp
@@ -19,6 +19,7 @@ void Ping360EthernetFinder::doBroadcast()
     // described in the communications manual.
     QByteArray datagram = "Discovery";
     _broadcastSocket.writeDatagram(datagram, QHostAddress::Broadcast, 30303);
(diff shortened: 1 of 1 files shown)


Williangalvani:
I expected Qt to broacast on all interfaces. I guess I was wrong.
Should we use something more generic instead?
There is an example [here](https://forum.qt.io/post/187963) of how to do it for all interfaces.

jaxxzer:
This works perfect and covers our use cases, the only time it won't work is if there is a third router/if on the computer. I don't think we need to worry about that. Manual connection works for any case.

patrickelectric:
I believe that an extra 4 lines to cover **all** cases does not hurt to add in this PR.

jaxxzer:
> I believe that an extra 4 lines to cover all cases does not hurt to add in this PR.

I haven't tested the link. I have tested this patch, it works perfectly to make our program work better, and it is ready to be merged with one click of a button..

PR #784, opened 2019-09-30
Fix PingComboSlider bind and reset settings

### LINKED ISSUE #781: Problems with sensor 'reset settings'
Reset settings is using the _firmware_ default range (1.5m) but the application uses a different default (2m). This causes the range to appear as (1.5m) on the display, and additionally, the slider gets out of sync:
Screenshot_20190926_154646

Also, I think that we should not modify the angle setting when reset settings is clicked, just the range, etc that the user controls in the panel.

patrickelectric:
Fix #781 

~@tcanabrava I was unable to fix the `value` bind that's probably lost here:~ https://github.com/bluerobotics/ping-viewer/blob/a1e1cb25e8a912161c9ebd056b88fad3d1d9201b/qml/PingComboSlider.qml#L67-L73

~After this, it's not possible to rebind the `value` property without binding loop or any crazy thing.
It was only possible with fe37f26, let me know if you thing that there is an alternative for that.~

Edit:

The actual version fix the issue, new PRs should be merged first and the commit should be organized better here.


src/sensor/ping360.cpp
```diff
@@ -498,15 +497,17 @@ void Ping360::resetSettings()
...
     _angle = _firmwareDefaultAngle;
-    _transmit_duration = _firmwareDefaultTransmitDuration;
-    _sample_period = _firmwareDefaultSamplePeriod;
-    _transmit_frequency = _firmwareDefaultTransmitFrequency;
-    _num_points = _firmwareDefaultNumberOfSamples;
-    deltaStep(_angle_offset, false);
+    // This variables should not be set directly without equality checks,
+    // such thing will result in a miss signal emition
+    set_gain_setting(_firmwareDefaultGainSetting);
+    set_transmit_duration(_firmwareDefaultTransmitDuration);
```
jaxxzer: This does not fully address #781 
There are two things remaining:
1. set the sonar to 2 meter range, not 1.5m (we want to use the _application_ default settings. not the _firmware_ default settings.)
2. delete line 501, do not update the angle

src/sensor/ping360.cpp
```diff
@@ -498,15 +497,17 @@ void Ping360::resetSettings()
...
+    set_sample_period(_firmwareDefaultSamplePeriod);
+    set_transmit_frequency(_firmwareDefaultTransmitFrequency);
+    set_number_of_points(_firmwareDefaultNumberOfSamples);
```
jaxxzer: We should be using 1200 points

tcanabrava:
> Fix #781
> 
> @tcanabrava I was unable to fix the `value` bind that's probably lost here:
> 
> https://github.com/bluerobotics/ping-viewer/blob/a1e1cb25e8a912161c9ebd056b88fad3d1d9201b/qml/PingComboSlider.qml#L67-L73
> 
> After this, it's not possible to rebind the `value` property without binding loop or any crazy thing.
> It was only possible with [fe37f26](https://github.com/bluerobotics/ping-viewer/commit/fe37f264cb5838659038708c7cc2c35794f7c88c), let me know if you thing that there is an alternative for that.

The problem is the `=`, it break the bindings. You need to use `Binding on` instead of the normal bindings for this to work.

    Binding on value {
        when: someItem.actionHappened
        value: someItem.value
    }
    onCompleted {
        value = bleh;
    }

patrickelectric:
@tcanabrava that is the problem, its a double/float property, I was unable to find a boolean that can 'indicates' when such value is able to set/get without resulting in a bind loop.

patrickelectric:
@tcanabrava any idea ?

jaxxzer:
ping @tcanabrava & @patrickelectric

tcanabrava:
overall I think we are hit some limitations on Qml and *probably* a full C++ item should be devised that combines sliders with combo boxes. but this works and fixes an issue.

qml/PingSlider.qml
```diff
@@ -10,14 +10,55 @@ RowLayout {
...
     property alias control: sliderControl
-    property var valueText: undefined
+    property var model: undefined
+    property var modelValue: undefined
+    property alias modelIndex: root.value
 
-    onValueChanged: {
-        spinBox.value = value
-        sliderControl.value = value
+    onModelValueChanged: {
```
jaxxzer: A little documentation on each step will be nice :)

patrickelectric: Done.

patrickelectric:
> It looks ok, unfortunately complicated but nice to have it working correctly.
> 
> To catch me up, can you explain your reasoning for combining both implementations into single pingslider qml?

The old implementation of pingcomboslider was much more complicated, this patch only updates PingSlider to provide the same functionality with less duplicated code and a unnecessary item.  This PingSlider was able to do the same thing with 20 lines.

I was impossible for me to fix PingSliderComboBox since it does not provide a clear/documented logic and the variable name was not helping. We can talk more about it in the next call. 
This new implementation fix some maintainability issues with the old implementation.

PR #794, opened 2019-10-07
Fix some todos

patrickelectric:


src/sensor/ping360.cpp
```diff
@@ -51,6 +51,7 @@ Ping360::Ping360()
     // QVector crashs when constructed in initialization list
     _data = QVector<double>(_maxNumberOfPoints, 0);
 
+    setName({"Ping360"});
```
jaxxzer: are the {} unnecessary?

patrickelectric: Nops, fixed.

PR #798, opened 2019-10-07
Fix todos:  encapsulate ping common variables

patrickelectric:
Encapsulate common ping variables inside a struct, this helps to understand what is sensor variables and what is not


patrickelectric:
Will fix the CI soon.

PR #806, opened 2019-10-09
Add bitrate information about link

### LINKED ISSUE #432: Show actual incoming profile frequency and incoming data bitrate
Show this info in the debug/ping status window. This will help us see if we are saturating the link, and if there are any discrepancies between target settings and actual results.

patrickelectric:
Fix #432

### DIFF
--- a/src/link/abstractlink.h
+++ b/src/link/abstractlink.h
@@ -250,6 +250,6 @@ protected:
 
 private:
     bool _autoConnect;
-    LinkType _type;
     QString _name;
+    LinkType _type;
 };
--- a/src/link/abstractlink.h
+++ b/src/link/abstractlink.h
@@ -2,6 +2,7 @@
 
 #include <QObject>
 #include <QTime>
+#include <QTimer>
 
 #include "linkconfiguration.h"
 
@@ -216,6 +217,26 @@ public:
         };
     }
 
+    /**
+     * @brief Download speed of link in bits
+     *
+     * @return float
+     */
+    float downSpeed()
+    {
+        return _bitRateDownSpeed.speed;
+    }
+
+    /**
+     * @brief Upload speed of link in bits
+     *
+     * @return float
+     */
+    float upSpeed()
+    {
+        return _bitRateUpSpeed.speed;
+    }
+
     Q_PROPERTY(qint64 byteSize READ byteSize NOTIFY byteSizeChanged)
     Q_PROPERTY(LinkConfiguration* configuration READ configuration NOTIFY configurationChanged)
     Q_PROPERTY(QTime elapsedTime READ elapsedTime NOTIFY elapsedTimeChanged)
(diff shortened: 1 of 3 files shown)


src/link/abstractlink.h
```diff
@@ -250,6 +250,6 @@ class AbstractLink : public QObject
 
 private:
     bool _autoConnect;
-    LinkType _type;
     QString _name;
```
jaxxzer: should _name be initialized in the constructor?

src/link/abstractlink.h
```diff
@@ -228,6 +249,8 @@ class AbstractLink : public QObject
     Q_PROPERTY(QTime totalTime READ totalTime NOTIFY totalTimeChanged)
     Q_PROPERTY(QString totalTimeString READ totalTimeString NOTIFY totalTimeChanged)
     Q_PROPERTY(AbstractLinkNamespace::LinkType type READ type WRITE setType NOTIFY linkChanged)
+    Q_PROPERTY(float upSpeed READ upSpeed NOTIFY linkChanged)
+    Q_PROPERTY(float downSpeed READ downSpeed NOTIFY linkChanged)
```
jaxxzer: `NOTIFY speedChanged`?

src/link/abstractlink.h
```diff
@@ -251,5 +275,22 @@ class AbstractLink : public QObject
...
+        float speed;
+        float numberOfBytes;
+
+        /**
+         * @brief Reset total number of bytes and set the current speed
+         *
+         */
+        void update()
+        {
+            speed = numberOfBytes;
```
jaxxzer: `speed = numberofBytes*8;`

maybe refactor `speed` -> `bitsPerSecond`

src/link/abstractlink.h
patrickelectric: Fixed.

src/link/abstractlink.h
patrickelectric: Yep, check: https://github.com/bluerobotics/ping-viewer/issues/808

src/link/abstractlink.h
patrickelectric: It would be better for us to work with bytes as a consistent unit, there is no advantage to use bits directly since our communication is bytes based and the conversion between both units are pretty simple. 
Also bits does not reflect directly in information, since our main unit for "data" are byte arrays from 0-255 for our sonars, it makes more sense to use bytes as a speed representation, otherwise we'll need to divide by 8 to understand what is going on.

patrickelectric: Let me know what you think.

jaxxzer: In my mind I would be comparing the number to the 115200 baud or 1Mbaud that we have configured to think about saturation, and the bandwidth measurements with iperf or companion to see the bandwidth comparision/contribution. I think you are right for your own reasons. It's nbd to convert.

PR #811, opened 2019-10-10
Force name in AbstractLink constructor 

### LINKED ISSUE #808: Force name argument in abstractlink constructor

patrickelectric:
We are helping ourselves forcing the name of the link in the constructor of an AbstractLink,
making it easier to provide the user the connection type that he's using. 
Fix #808


patrickelectric:
> Should we remove reference to AbstractLink::setType()?

It's possible that we could create abstract classes with specific names but without knowing the type of the connection, so it's better to let `setType` exist for future code, and it helps to make the constructor simpler with only the name as argument.

PR #813, opened 2019-10-14
Random fixes corrections around

patrickelectric:
- Move util function to static and fix shadow variable
- Add const reference for some parameters
- Remove unnecessary constructor 
- Use QRandomGenerator over qrand

### DIFF
--- a/src/link/ping1dsimulationlink.cpp
+++ b/src/link/ping1dsimulationlink.cpp
@@ -35,12 +35,12 @@ void Ping1DSimulationLink::randomUpdate()
     for (int i = 0; i < numPoints; i++) {
         float point;
         if (i < stop1) {
-            point = 0.1 * (qrand() % 256);
+            point = 0.1 * randomPoint<uint8_t>();
         } else if (i < stop2) {
             point
                 = 255 * ((-4.0 / qPow((stop2 - stop1), 2.0)) * qPow((i - stop1 - ((stop2 - stop1) / 2.0)), 2.0) + 1.0);
         } else {
-            point = 0.45 * (qrand() % 256);
+            point = 0.45 * randomPoint<uint8_t>();
         }
         profile.set_profile_data_at(i, point);
     }
--- a/src/link/linkconfiguration.h
+++ b/src/link/linkconfiguration.h
@@ -54,8 +54,8 @@ public:
      * @param name
      * @param deviceType
      */
-    LinkConfiguration(LinkType linkType = LinkType::None, QStringList args = QStringList(), QString name = QString(),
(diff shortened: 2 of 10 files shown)


src/filemanager/filemanager.h
```diff
@@ -111,7 +111,7 @@ class FileManager : public QObject
      *
      */
     struct FolderStruct {
-        FolderStruct(QDir _dir, QString _extension = QString())
+        FolderStruct(QDir _dir, const QString& _extension = QString())
```
tcanabrava: I'd remove that completely. FolderStruct {dir, extension}; already creates what you want.

patrickelectric: LCMTH, great catch

src/link/ping1dsimulationlink.cpp
```diff
@@ -35,12 +36,12 @@ void Ping1DSimulationLink::randomUpdate()
     for (int i = 0; i < numPoints; i++) {
         float point;
         if (i < stop1) {
-            point = 0.1 * (qrand() % 256);
+            point = 0.1 * QRandomGenerator::global()->bounded(256);
```
Williangalvani: why 256? couldn't you do `QRandomGenerator::global()->bounded(25.6)`?

src/link/ping1dsimulationlink.cpp
```diff
@@ -35,12 +36,12 @@ void Ping1DSimulationLink::randomUpdate()
...
         } else if (i < stop2) {
             point
                 = 255 * ((-4.0 / qPow((stop2 - stop1), 2.0)) * qPow((i - stop1 - ((stop2 - stop1) / 2.0)), 2.0) + 1.0);
         } else {
-            point = 0.45 * (qrand() % 256);
+            point = 0.45 * QRandomGenerator::global()->bounded(256);
```
Williangalvani: same here

src/link/ping1dsimulationlink.cpp
patrickelectric: From the doc: 
`Generates one random double in the range between 0 (inclusive) and highest (exclusive).`
Probably you'll catch it better as an [0, 256) interval that is the same as [0, 255].

patrickelectric: @Williangalvani let me know if this resolve the conversation

Williangalvani: The code says the function is equivalent to `generateDouble() * highest`, so I thought you just move that 0.1 inside. I understand that double and integer generation will yeld different results, but I thought it didn't matter as much here.

I'm not really concerned about this, though, marking as resolved.

jaxxzer:
lgtm

PR #815, opened 2019-10-14
pingviewer: Use fast compilation flags in release mode

patrickelectric:


patrickelectric:
Some variables are optimized with fast flags, removing such flags in debug more solves that problem

PR #820, opened 2019-11-07
Fix singleton self call

### LINKED ISSUE #819: Remove memory leak from singletons self calls

patrickelectric:
Fix #819

### DIFF
--- a/src/devicemanager/devicemanager.cpp
+++ b/src/devicemanager/devicemanager.cpp
@@ -1,3 +1,5 @@
+#include <QQmlEngine>
+
 #include "devicemanager.h"
 #include "filelink.h"
 #include "logger.h"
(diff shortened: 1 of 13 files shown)


src/devicemanager/devicemanager.cpp
```diff
@@ -9,6 +11,8 @@ PING_LOGGING_CATEGORY(DEVICEMANAGER, "ping.devicemanager");
 DeviceManager::DeviceManager()
     : _detector(new ProtocolDetector())
 {
+    QQmlEngine::setObjectOwnership(this, QQmlEngine::CppOwnership);
```
jaxxzer: can we get a comment? maybe it is common qt programming practice, but I don't know the reason for this addition.

patrickelectric: The idea here is to set the ownership of this class as cpp and not javascript, as documented here: https://doc.qt.io/qt-5/qqmlengine.html#ObjectOwnership-enum
The documentation is pretty explicit about what is going on here, we don't want this classes to be deleted by our QML engine but for the cpp scope.

jaxxzer: Thank you :+1: I did look on the qt doc, but did not find that enum on my own.

PR #821, opened 2019-11-11
devicemanager: Fix ownership of devicemanager

patrickelectric:


patrickelectric:
The problem is that the deviceManager has a list of Sensors that are the parent of the visualization widgets resulting in the problem described of double freedom.

PR #822, opened 2019-11-11
Save user manual connections to be scanned 

### LINKED ISSUE #818: Remember Manual Connections Across App Restarts
## Summary

As I understand, Ping-Viewer only looks at 192.168.2.2 for connecting to PingProxy automatically. If the device running PingProxy has a different address, you have to enter it manually every time Ping-Viewer is started. 

It would be nice if there were the ability to remember manually entered configurations across restarts of the app. Perhaps only save them if we receive a valid Ping packet in return. Alternatively, there could be a button in the settings panel to save a connection as a 'favorite'.

Ideally, these saved connections or *favorites* would be scanned when the app is restarted so that the online devices populate in the device manager, but devices that are off/not responding don't show up. 

## Additional information

There are lots of situations where Ping devices may be on various different hosts, not just a standard BlueROV2. For example, if you're connected to an ROV and a surface vessel at the same time. Or multiple ROVs. Or maybe you have 10 buoys on a lake and each of them has a Pi and Ping.

patrickelectric:
Fix #818 
out

### DIFF
--- a/src/sensor/ping.cpp
+++ b/src/sensor/ping.cpp
@@ -83,11 +83,6 @@ Ping::Ping()
                 "ping1d", std::bind(&Ping::checkNewFirmwareInGitHubPayload, this, std::placeholders::_1));
         }
     });
-
-    // TODO: @Patrick move it to devicemanager
-    // Load last successful connection
-    auto config = SettingsManager::self()->lastLinkConfiguration();
-    qCDebug(PING_PROTOCOL_PING) << "Loading last configuration connection from settings:" << config;
 }
 
 void Ping::startPreConfigurationProcess()
@@ -97,8 +92,6 @@ void Ping::startPreConfigurationProcess()
         qCDebug(PING_PROTOCOL_PING) << "It's only possible to set last configuration when link is writable.";
         return;
     }
-
-    SettingsManager::self()->lastLinkConfiguration(*link()->configuration());
 
     // Request device information
     request(Ping1dId::PING_ENABLE);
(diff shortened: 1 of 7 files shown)


src/sensor/ping.cpp
```diff
@@ -103,8 +103,6 @@ void Ping::startPreConfigurationProcess()
 
     // Start periodic request timer
     _periodicRequestTimer.start();
-
-    // Save configuration
```
jaxxzer: squash with parent?

src/settings/settingsmanager.cpp
```diff
@@ -17,10 +18,45 @@ SettingsManager::SettingsManager()
...
+            }),
+        _lastLinkConfigurations.end());
+
+    // Move to QVariantList before saving
+    // This is necessary since QList<QDataStream operator> does not work as expected
+    QVariantList variants;
+    for (const auto linkConfiguration : _lastLinkConfigurations) {
+        variants.append(QVariant::fromValue(linkConfiguration));
+    }
+    set("lastLinkConfigurations", variants);
```
jaxxzer: why use `set` instead of `=`?

src/settings/settingsmanager.cpp
```diff
@@ -17,10 +18,45 @@ SettingsManager::SettingsManager()
...
+    for (const auto variant : variantLastLinkConfigurations) {
+        _lastLinkConfigurations.append(variant.value<LinkConfiguration>());
+    }
+}
+
+void SettingsManager::saveLinkConfigurations()
+{
+    // Remove duplicated before saving it
+    _lastLinkConfigurations.erase(
+        std::unique(_lastLinkConfigurations.begin(), _lastLinkConfigurations.end(),
```
jaxxzer: nice

src/settings/settingsmanager.cpp
patrickelectric: `Set` check if the variable exist or not, print a debug message related to the variable existence, the old and new value of such variable and also forces the sync of the settings.

PR #823, opened 2019-11-12
Add ping360 speed test and some new features 

### LINKED ISSUE #796: Add test for scan speed
Add test to maintain times #789 valid

patrickelectric:
- Add command line parser 
- Add writable ping360 simulation
- Add Ping360 profile request speed test

Fix #796

### DIFF
--- a/qml/DeviceManagerViewer.qml
+++ b/qml/DeviceManagerViewer.qml
@@ -16,7 +16,18 @@ PingPopup {
     }
 
     Component.onCompleted: {
-        open()
+        if(!DeviceManager.primarySensor) {
+            open()
+        }
+    }
+
+    Connections{
+        target: DeviceManager
+        onPrimarySensorChanged: {
+            if(DeviceManager.primarySensor) {
+                close()
+            }
+        }
     }
 
     ColumnLayout {
@@ -171,9 +182,6 @@ PingPopup {
                         hoverEnabled: true
                         onClicked: {
                             DeviceManager.connectLink(connection)
-                            // We should check if Sensor device is created and if it's really connected and
-                            // communicating with the device
-                            root.close()
                         }
                     }
 
--- a/src/commandline/commandlineparser.h
+++ b/src/commandline/commandlineparser.h
@@ -1,0 +1,46 @@
+#pragma once
+
+#include "devicemanager.h"
+#include <QCommandLineParser>
+#include <QLoggingCategory>
+
+Q_DECLARE_LOGGING_CATEGORY(COMMANDLINEPARSER)
+
+/**
+ * @brief Deal with command lines
+ *
+ */
+class CommandLineParser : public QCommandLineParser {
+public:
+    /**
+     * @brief Construct a new Command Line Parser object
+     *  The constructor will deal with the arguments
+     *
+     * @param app
+     */
+    CommandLineParser(const QCoreApplication& app);
+
+    /**
+     * @brief Destroy the Command Line Parser object
+     *
+     */
+    ~CommandLineParser() = default;
+
+private:
+    struct OptionStruct {
+        QCommandLineOption option;
+        std::function<void(const QString&)> function;
+    };
+
+    QList<OptionStruct> _optionsStruct {
+        {
+            {{"connect", "c"}, "Connect directly with the input.", "connectionString"
+#if defined(PING360_SPEED_TEST)
+                ,
+                "Ping360:6"
+#endif
+            },
+            [](const QString& result) { DeviceManager::self()->connectLinkDirectly(LinkConfiguration {result}); },
+        },
+    };
+};
--- a/src/devicemanager/devicemanager.h
+++ b/src/devicemanager/devicemanager.h
@@ -59,7 +59,14 @@ public:
      *
      * @param linkConf
      */
-    Q_INVOKABLE void connectLink(LinkConfiguration* linkConf);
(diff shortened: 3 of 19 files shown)


jaxxzer:
> Fix #789

789 is a merged pr. is that correct?

patrickelectric:
> > Fix #789
> 
> 789 is a merged pr. is that correct?

No, the correct issue is #796, thanks for pointing.

src/commandline/commandlineparser.h
```diff
@@ -0,0 +1,46 @@
...
+     */
+    ~CommandLineParser() = default;
+
+private:
+    struct OptionStruct {
+        QCommandLineOption option;
+        std::function<void(const QString&)> function;
+    };
+
+    QList<OptionStruct> _optionsStruct {
```
tcanabrava: QVector

src/link/ping360simulationlink.cpp
```diff
@@ -45,5 +65,30 @@ void Ping360SimulationLink::randomUpdate()
...
+    }
+
+#if defined(PING360_SPEED_TEST)
+    if (_spins > simulationTest.spins) {
+        // Since we are forcing the application to close, qDebug may not work
+        printf("%s: Average elapsed request elapsed time %f [Valid: %f]\n", __PRETTY_FUNCTION__, _globalAverageTimeMs,
+            simulationTest.maxAverageRequestTimeMs);
+
+        if (_globalAverageTimeMs > simulationTest.maxAverageRequestTimeMs) {
+            QCoreApplication::exit(-1);
```
tcanabrava: use proper error codes from errno.h ?

src/commandline/commandlineparser.h
patrickelectric: does not compile.

src/devicemanager/devicemanager.h
```diff
@@ -59,7 +59,14 @@ class DeviceManager : public QAbstractListModel {
      *
      * @param linkConf
      */
-    Q_INVOKABLE void connectLink(LinkConfiguration* linkConf);
+    Q_INVOKABLE void connectLink(const LinkConfiguration* linkConf);
```
jaxxzer: Is there a reason this should not be passed by reference like the other?

qml/DeviceManagerViewer.qml
```diff
@@ -16,7 +16,18 @@ PingPopup {
     }
 
     Component.onCompleted: {
-        open()
+        if(!DeviceManager.primarySensor) {
```
jaxxzer: Will this interfere with manual connections where the sensor does not immediately respond? ie. if you wanted to connect manually to /dev/ttyUSB0 to do a ping1d recovery with the stm32 bootloader.

src/commandline/commandlineparser.h
```diff
@@ -34,7 +34,12 @@ class CommandLineParser : public QCommandLineParser {
 
     QList<OptionStruct> _optionsStruct {
         {
-            {{"connect", "c"}, "Connect directly with the input.", "connectionString"},
+            {{"connect", "c"}, "Connect directly with the input.", "connectionString"
```
jaxxzer: squash with parent?

src/link/ping360simulationlink.cpp
```diff
@@ -66,11 +68,27 @@ void Ping360SimulationLink::randomUpdate()
...
+    _globalAverageTimeMs = _globalAverageTimeMs * weight + elapsedTime * (1 - weight);
+
     _elapsedTimer.restart();
+
+    // This should be done after _counter increase to not interate before a full first spin
+    if (_counter % angularResolution == 0) {
+        _spins++;
+    }
+
+#if defined(PING360_SPEED_TEST)
```
jaxxzer: simulationlink seems like the wrong place to perform all this elapsed time calculations and code for the test.

I would put this in sensor like Ping360::doFullScanSpeedTest().

src/link/ping360simulationlink.cpp
```diff
@@ -45,5 +63,14 @@ void Ping360SimulationLink::randomUpdate()
     deviceData.updateChecksum();
     emit newData(QByteArray(reinterpret_cast<const char*>(deviceData.msgData), deviceData.msgDataLength()));
 
-    counter++;
+    // Calculate the global average time between requests
```
jaxxzer: This hunk (and related declarations above) belong with e845f3c031e5f0404e8e075eef4d564a3b95ec07

src/devicemanager/devicemanager.h
patrickelectric: this function s called from the QML side, where QObjects are only used with pointers.

qml/DeviceManagerViewer.qml
patrickelectric: No, primary sensor is a pointer for the connected device, if there is something connected the is valid.

src/commandline/commandlineparser.h
patrickelectric: Done.

src/link/ping360simulationlink.cpp
patrickelectric: This will not simulate the link connection and everything related to the latency between sensor and the abstractlink logic.

patrickelectric: I agree that is not **the best** place to have it, but we can improve it later.

src/link/ping360simulationlink.cpp
patrickelectric: Done.

PR #824, opened 2019-11-13
ping360: Fix reset feature

### LINKED ISSUE #781: Problems with sensor 'reset settings'

patrickelectric:
This commit adds a tracking interface for the sensor settings,
making it possible to invalidate incoming messages and force the new
settings.

Fix #781

### DIFF
--- a/src/sensor/ping360.cpp
+++ b/src/sensor/ping360.cpp
@@ -261,7 +261,10 @@ void Ping360::handleMessage(const ping_message& msg)
         // This properties are changed internally only when the link is not writable
         // Such information is normally sync between our application and the sensor
         // So with normal links such attribution is not necessary
-        if (!link()->isWritable()) {
+        _sensorSettings.checkValidation({deviceData.transmit_duration(), deviceData.gain_setting(),
+            deviceData.data_length(), deviceData.sample_period(), deviceData.transmit_frequency()});
+        // Everything should be valid, otherwise the sensor is not in sync
+        if (!link()->isWritable() && _sensorSettings.valid) {
             set_gain_setting(deviceData.gain_setting());
             set_transmit_duration(deviceData.transmit_duration());
             set_sample_period(deviceData.sample_period());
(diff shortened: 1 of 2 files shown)


src/sensor/ping360.cpp
```diff
@@ -484,7 +486,7 @@ uint16_t Ping360::calculateSamplePeriod(float distance)
                                        << "distance:" << distance << "_num_points" << _num_points << "_speed_of_sound"
                                        << _speed_of_sound << "_samplePeriodTickDuration" << _samplePeriodTickDuration;
 
-        return _firmwareDefaultSamplePeriod;
+        return _viewerDefaultSamplePeriod;
```
jaxxzer: https://github.com/bluerobotics/ping-viewer/pull/824/commits/cdfd858fe49607145ed21d7386092c47c3bcb735#diff-10710c12df3d3ecf2ad7777360a726b5R484
'going to use viewer default value'

Side Note: I would put an assert here instead of warning/return. This should never happen (only valid distance values should be submitted for arguments).

patrickelectric: An assert will finish the application, making the user angry. 
The application can run and exit if the user would like to do so, If the intended behavior is not clear, he can send us a log and we can catch with the warning message.
Imagine that you are the user of our application in the middle of the ocean without internet connection and your application starts to close because we have a miss-configuration somewhere. Probably he'll be very angry that the application was killing itself because of a fail test that is not critical for the application to work. 
Is like closing the application if the interface color does not match with the settings, it's not a important thing and the user can still use the interface and deal with it somehow, but closing the application only because it's not following our expectation is something that I would like to avoid as much as possible in our software, this is not the intended behavior of an **assert**.

patrickelectric: Also take in mind that this PR does not change how the logic works, only the value that will return if something wrong happens. So the assert discussion does not make sense in this PR since the work is based in the original implementation.

jaxxzer: > Also take in mind that this PR does not change how the logic works, only the value that will return if something wrong happens. So the assert discussion does not make sense in this PR since the work is based in the original implementation.

:+1:  That is why I discussed it as a 'side note' (a related topic for another discussion)

jaxxzer: > this is not the intended behavior of an assert.

I will argue that this is exactly the intended behavior of an assert. Asserts can also be made no-op in release mode.

Assert is used to catch bugs in code that should never execute or conditions that should never be true. This code should never execute. We should add an assert to catch early in dev and fix our program to ensure that this code never executes. We should not return a bogus value.

Like you said, we can fix it later.

PR #826, opened 2019-11-14
Improve chart resolution

### LINKED ISSUE #732: L5: verify ascan resolution
> The Ascan amplitude display looks smoothed compared to my original, have
you cut the number of points down or averaged?

patrickelectric:
I did a quick text with 1k points, as you can check here the resolution matchs:
image

My display has 1080 pixels.

Fix #732


jaxxzer:
This works with ping1d also?

patrickelectric:
> This works with ping1d also?

Yes.

PR #827, opened 2019-11-15
Fix PingSlider ruler background

### LINKED ISSUE #787: Sliders does not follow the same style
image

### LINKED ISSUE #788: Use odd ticks in sliders
image

patrickelectric:
Remove SliderRuler code and use simple background item.
This results in a much simpler code base [-300 lines] and the same style/size between sliders, making
the interface more uniform. 

With the PR:
image

Master:
image

Fix #787 and Fix #788


qml/PingSlider.qml
```diff
@@ -4,15 +4,14 @@ import QtQuick.Controls.Styles 1.0
...
-import SliderRuler 1.0
-
 RowLayout {
     id: root
 
     property var text
     property double value: 0
     property double from: 0
     property double to: 0
+    property double ticksNumber: 6
```
Williangalvani: what about just making it 7 instead of using `ticksNumber + 1` in the model? We do get 7 ticks after all.

patrickelectric: This property allow us to change the ticks number, the model will always be "ticks number" + 1;

PR #835, opened 2019-11-26
Add gradient scale

### LINKED ISSUE #391: Add a gradient scale
Add a gradient scale to the corner/edge of the display showing the current gradient for the waterfall

patrickelectric:
Fix #391 

image

image


patrickelectric:
@jaxxzer is not possible to create dynamic gradients in QML with different stops or number of stops, it's also not possible to access the stop list from C++ and vice-versa.
If you can please provide a full example that replicates the behavior that this PR, working for each different theme in our waterfallgradient and also dynamic enough to change for each theme change.

jaxxzer:
that's ok, just merge this when you are ready

PR #848, opened 2019-12-05
Add mavlinkmanager and heading logic for ping360

patrickelectric:


src/devicemanager/devicemanager.h
```diff
@@ -59,7 +59,7 @@ class DeviceManager : public QAbstractListModel {
      *
      * @param linkConf
      */
-    Q_INVOKABLE void connectLink(const LinkConfiguration* linkConf);
+    Q_INVOKABLE void connectLink(LinkConfiguration* linkConf);
```
tcanabrava: this is also in another pr, no?

src/link/udplink.cpp
```diff
@@ -13,7 +13,20 @@ UDPLink::UDPLink(QObject* parent)
...
+    connect(_udpSocket, &QIODevice::readyRead, this, [this] { emit newData(_udpSocket->readAll()); });
+    connect(_udpSocket, QOverload<QAbstractSocket::SocketError>::of(&QAbstractSocket::error), this,
+        [this](QAbstractSocket::SocketError /*socketError*/) { printErrorMessage(); });
+
+    // QUdpSocket fail to emit state signal
+    // Here we use a timer to check if we are in a connect state, if not we try again
+    connect(&_stateTimer, &QTimer::timeout, this, [this] {
+        if (_udpSocket->state() == QAbstractSocket::UnconnectedState) {
+            printErrorMessage();
+            qDebug(PING_PROTOCOL_UDPLINK) << "Trying to connect with host again.";
```
tcanabrava: trying to reconnect with host?

src/mavlink/mavlinkmanager.cpp
```diff
@@ -0,0 +1,80 @@
...
+    connect({LinkType::Udp, {"192.168.2.2", "14660"}});
+
+    QObject::connect(&_heartbeatTimer, &QTimer::timeout, this, &MavlinkManager::sendHeartbeatMessage);
+    _heartbeatTimer.start(4000);
+}
+
+void MavlinkManager::connect(const LinkConfiguration& conf)
+{
+    if (_linkIn) {
+        auto link = _linkIn->self();
```
tcanabrava: the method `self` is quite strange, it sounds like `this` as it's the same thing that python uses.

src/mavlink/mavlinkmanager.cpp
```diff
@@ -0,0 +1,80 @@
...
+        if (mavlink_parse_char(0, byte, &message, &status)) {
+            emit mavlinkMessage(message);
+        }
+    }
+}
+
+void MavlinkManager::sendHeartbeatMessage()
+{
+    static mavlink_message_t message;
+    static unsigned char bufffer[1024];
```
tcanabrava: I wonder if the `static thread mutex` that you discovered will be bad here. maybe those should be defined in an unamed namespace in the cpp context?

src/devicemanager/devicemanager.h
patrickelectric: Yes, I'll rebase over master.

src/link/udplink.cpp
patrickelectric: Good.

src/mavlink/mavlinkmanager.cpp
patrickelectric: That's a good idea, but I would prefer to initialize the message in compile time.

patrickelectric: Fixed.

src/mavlink/mavlinkmanager.cpp
patrickelectric: Indeed, we need to rename it. Link is a helper object for the AbstractLink pointer, we can rework that in the future.

PR #853, opened 2019-12-09
ping360: Use double as argument

patrickelectric:
This makes more sense since the PROPERTY is double and the range() return is also double


patrickelectric:
@tcanabrava yes, it's the QML, that's why I'm making the property to match the function.

PR #866, opened 2019-12-16
ping: Move from QThread::msleep to QTimer::singleShot

patrickelectric:
This approach will fix interface hiccups while flashing
avoiding the sleep of the main thread

### DIFF
--- a/src/sensor/ping.cpp
+++ b/src/sensor/ping.cpp
@@ -358,26 +358,30 @@ void Ping::flash(const QString& fileUrl, bool sendPingGotoBootloader, int baud, 
     }
 
     qCDebug(PING_PROTOCOL_PING) << "Finish connection.";
-    // TODO: Move thread delay to something more.. correct.
-    QThread::msleep(1000);
-    link()->finishConnection();
 
-    QSerialPortInfo pInfo(serialLink->port()->portName());
-    QString portLocation = pInfo.systemLocation();
+    auto flashSensor = [=] {
+        flasher()->setBaudRate(baud);
+        flasher()->setFirmwarePath(fileUrl);
+        flasher()->setLink(link()->configuration()[0]);
+        flasher()->setVerify(verify);
+        flasher()->flash();
+    };
 
-    qCDebug(PING_PROTOCOL_PING) << "Save sensor configuration.";
-    updatePingConfigurationSettings();
+    auto finishConnection = [=] {
+        link()->finishConnection();
 
-    qCDebug(PING_PROTOCOL_PING) << "Start flash.";
-    QThread::msleep(1000);
+        QSerialPortInfo pInfo(serialLink->port()->portName());
+        QString portLocation = pInfo.systemLocation();
 
-    flasher()->setBaudRate(baud);
-    flasher()->setFirmwarePath(fileUrl);
-    flasher()->setLink(link()->configuration()[0]);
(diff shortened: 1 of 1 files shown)


src/sensor/ping.cpp
```diff
@@ -359,25 +359,33 @@ void Ping::flash(const QString& fileUrl, bool sendPingGotoBootloader, int baud,
 
     qCDebug(PING_PROTOCOL_PING) << "Finish connection.";
     // TODO: Move thread delay to something more.. correct.
-    QThread::msleep(1000);
-    link()->finishConnection();
+    auto flashSensor = std::function<void(void)>();
```
tcanabrava: auto flashSensor = [] { ... } would work the same?

src/sensor/ping.cpp
```diff
@@ -359,25 +359,33 @@ void Ping::flash(const QString& fileUrl, bool sendPingGotoBootloader, int baud,
...
+    auto finishConnection = std::function<void(void)>();
+
+    flashSensor = [=] {
```
jaxxzer: sort of what tomaz said, why don't you just declare + define in one line?

src/sensor/ping.cpp
```diff
@@ -359,25 +359,33 @@ void Ping::flash(const QString& fileUrl, bool sendPingGotoBootloader, int baud,
...
+        flasher()->setBaudRate(baud);
+        flasher()->setFirmwarePath(fileUrl);
+        flasher()->setLink(link()->configuration()[0]);
+        flasher()->setVerify(verify);
+        flasher()->flash();
+    };
+
+    finishConnection = [=] {
+        QThread::msleep(1000);
```
jaxxzer: should this line be here? it looks like we are still sleeping the main thread.

src/sensor/ping.cpp
```diff
@@ -359,25 +359,33 @@ void Ping::flash(const QString& fileUrl, bool sendPingGotoBootloader, int baud,
...
-    flasher()->setBaudRate(baud);
-    flasher()->setFirmwarePath(fileUrl);
-    flasher()->setLink(link()->configuration()[0]);
-    flasher()->setVerify(verify);
-    flasher()->flash();
+        QTimer::singleShot(1000, flashSensor);
+    };
+    QTimer::singleShot(1000, finishConnection);
 
-    QThread::msleep(500);
```
jaxxzer: Is there anything wrong with sleeping the main thread? Can you explain what the hiccups in the interface is?

tcanabrava: you should never sleep the main thread :) it hangs the QEventLoop

src/sensor/ping.cpp
patrickelectric: Because `finishConnection` needs `flashSensor` and `finishConnection` will be called before `flashSensor`. Doing the declaration of each lambda inside the function will break the readability since `finishConnection`  happens before `flashSensor` but the declaration not.

src/sensor/ping.cpp
patrickelectric: Checks my previous answer: 
https://github.com/bluerobotics/ping-viewer/pull/866#discussion_r358879255

src/sensor/ping.cpp
patrickelectric: lost code, will be removed.

PR #875, opened 2020-01-09
faq/troubleshooting: add some additional troubleshooting steps

jaxxzer:


jaxxzer:
ci fails because our site is down today

PR #884, opened 2020-02-14
Use struct to handle group of messages

patrickelectric:
This is important to know if the message was received and correct initialized


patrickelectric:
@tcanabrava common variables are the common variables inside the ping protocol scope.
[Some messages exist under some specific messages](https://docs.bluerobotics.com/ping-protocol/pingmessage-common/), that's why I have created this next struct level.
This is not doing much now, but in a future PR I'll need to verify if this specific variables set are initialized or not (this is a result of old firmwares and legacy behaviour), besides of the async nature of the messages/sensor. 
I would like to organize this in a better way if possible, and somehow with a initialization field that could help the child class (ping and ping360) to identify if this variables are updated and available to be used.

PR #886, opened 2020-02-14
ping360: Use designator in initialization

patrickelectric:


tcanabrava:
but I don't think this works correctly in all compilers yet, let's check the error messages.

patrickelectric:
appveyor does not have Qt in MSVC 2019 image.

PR #891, opened 2020-02-18
Ping360: Add checkbox for heading integration

### LINKED ISSUE #890: Add option to remove ping360 heading integration
From @kklemens:
> Ping-Viewer t2.0.6
Recommendations:
Have an option to turn off the Ping360 chart rotation if someone doesn’t want it. We have some “old school” customers.

patrickelectric:
Fix #890

### DIFF
--- a/qml/Ping360Visualizer.qml
+++ b/qml/Ping360Visualizer.qml
@@ -379,9 +379,22 @@ Item {
                 onCurrentTextChanged: waterfall.theme = currentText
             }
 
+            CheckBox {
+                id: headingIntegration
+                text: "Heading integration"
+                checked: true
+                Layout.columnSpan: 5
+                Layout.fillWidth: true
+                Layout.minimumWidth: 200
+                onCheckStateChanged: {
+                    ping.enableHeadingIntegration(checkState)
+                }
+            }
+
             Settings {
                 category: "Ping360Visualizer"
                 property alias flipAScanState: flipAScan.checkState
+                property alias headingIntegrationCheckState: headingIntegration.checkState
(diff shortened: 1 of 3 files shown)


src/sensor/ping360.cpp
```diff
@@ -531,4 +526,23 @@ void Ping360::resetSettings()
...
+        disconnect(MavlinkManager::self(), &MavlinkManager::mavlinkMessage, this, &Ping360::processMavlinkMessage);
+        _heading = 0;
+        emit headingChanged();
+    }
+}
+
+void Ping360::processMavlinkMessage(const mavlink_message_t& message)
+{
+    mavlink_attitude_t attitude;
+    mavlink_msg_attitude_decode(&message, &attitude);
```
jaxxzer: I think we should check the message type of `message` or change the type of the parameter to `const mavlink_attitude_t&`

qml/Ping360Visualizer.qml
```diff
@@ -379,9 +379,22 @@ Item {
                 onCurrentTextChanged: waterfall.theme = currentText
             }
 
+            CheckBox {
+                id: headingCorrection
```
jaxxzer: nitpick: it is funny to have two terms for the same thing: heading correction and heading integration

src/sensor/ping360.cpp
patrickelectric: We only decode attitude messages, but done.

qml/Ping360Visualizer.qml
patrickelectric: Done.

PR #898, opened 2020-02-20
protocoldetector: lengthen line break for ping1d

jaxxzer:


patrickelectric:
@tcanabrava this patch is for a hardware communication with a defined time between operations for proper hardware configuration.
A solution for this is the creation of a thread to do the necessary probe and hardware configuration, avoiding any sleep in the main thread. but this in unrelated to this patch.

About the increase of the sleep call, we need to help the serial break condition longer for proper hardware identification.

PR #905, opened 2020-03-04
Ping360 async

patrickelectric:
Add support to Ping360 async messages
- It only runs with our last test version 0.0.3
- We need to update the firmware check values to match our firmware release version
    - We can do that after testing it with out last test firmware and approving this patch

### DIFF
--- a/src/sensor/ping360.cpp
+++ b/src/sensor/ping360.cpp
@@ -97,6 +97,32 @@ Ping360::Ping360()
         emit messageFrequencyChanged();
     });
 
+    connect(this, &Ping360::firmwareVersionMinorChanged, this, [this] {
+        qCDebug(PING_PROTOCOL_PING360) << "Firmware version:"
+                                       << _commonVariables.deviceInformation.firmware_version_major
+                                       << _commonVariables.deviceInformation.firmware_version_minor
+                                       << _commonVariables.deviceInformation.firmware_version_patch;
+
+        // Wait for firmware information to be available before looking for new versions
+        static bool once = false;
+        if (!once && _commonVariables.deviceInformation.initialized) {
+            once = true;
+
+            if (_commonVariables.deviceInformation.firmware_version_major == 3
+                && _commonVariables.deviceInformation.firmware_version_minor == 3
+                && _commonVariables.deviceInformation.firmware_version_patch == 1) {
+                _profileRequestLogic.type = Ping360RequestStateStruct::Type::AutoTransmitAsync;
+            } else {
+                _profileRequestLogic.type = Ping360RequestStateStruct::Type::Legacy;
+            }
+        }
+    });
+
+    connect(this, &Ping360::sectorSizeChanged, this, [this] {
+        _sensorSettings.start_angle = angle_offset() - _sectorSize / 2;
+        _sensorSettings.end_angle = (angle_offset() + _sectorSize / 2 - 1) % 400;
+    });
+
     // By default heading integration is enabled
     enableHeadingIntegration(true);
 }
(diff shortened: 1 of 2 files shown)


Williangalvani:
I did some quick tests here: 
 - It seems to fail at any distance other than 2 meters (stays still vibrating?)
 - Can we make it stop when we close ping-viewer?

patrickelectric:
It appears that [actual auto_transmit message](https://github.com/bluerobotics/ping-protocol/blob/a83fba61a17920c6736b5c92a6b88e2128f59d70/src/definitions/ping360.json#L226)
is a bad message to configure the sensor, since it does not allow as much configuration as [transducer_message](https://github.com/bluerobotics/ping-protocol/blob/master/src/definitions/ping360.json#L172).
I'll continue my tests, but with the actual structure, some logic will be necessary to deal with this, and  will probably not be user friendly if someone wants to use our api.

src/sensor/ping360.cpp
```diff
@@ -548,4 +662,24 @@ void Ping360::processMavlinkMessage(const mavlink_message_t& message)
...
+    if (_profileRequestLogic.type == Ping360RequestStateStruct::Type::Legacy) {
+        return messageFrequencies[Ping360Id::DEVICE_DATA].frequency;
+    }
+    return messageFrequencies[Ping360Id::AUTO_DEVICE_DATA].frequency;
+}
+
+Ping360::~Ping360() {
+    updateSensorConfigurationSettings();
+
+    // TODO: Find a better way
```
jaxxzer: I think line break may work.

src/sensor/ping360.h
```diff
@@ -617,6 +622,9 @@ class Ping360 : public PingSensor {
             valid = other.transmit_duration == transmit_duration && other.gain_setting == gain_setting
                 && other.num_points == num_points && other.sample_period == sample_period
                 && other.transmit_frequency == transmit_frequency;
+            if (checkSector && valid) {
+                valid = valid && other.start_angle == start_angle && other.end_angle == end_angle;
```
jaxxzer:
```suggestion
                valid = other.start_angle == start_angle && other.end_angle == end_angle;
```
valid must be true, it is redundant to do the comparison again.

src/sensor/ping360.cpp
patrickelectric: It'll fail via UDP communication.

src/sensor/ping360.h
patrickelectric: Good point :+1:

patrickelectric:
@Williangalvani can you test with the last test version 0.0.3 ?

Williangalvani:
> @Williangalvani can you test with the last test version 0.0.3 ?

Worked fine on the bench, I'll test it in the ROV later today

src/sensor/ping360.cpp
jaxxzer: aha

Williangalvani:
I got almost twice the speed with the new fw and this PR, nice job!
66hz now from 35hz before @2m

src/sensor/ping360.cpp
```diff
@@ -283,6 +350,60 @@ void Ping360::handleMessage(const ping_message& msg)
...
+
+        // Restart timer, if the channel allows it
+        if (link()->isWritable()) {
+            // Use 200ms for network delay
+            const int profileRunningTimeout = _angular_speed / _angularSpeedGradPerMs + 200;
+            _timeoutProfileMessage.start(profileRunningTimeout);
+        }
+
+        _data.resize(autoDeviceData.data_length());
+#pragma omp for
```
tcanabrava, line 368: move the pragma to the same line as the for

tcanabrava: while I do really - *really* - like openmp, perhaps we should note that it does not have a really good suppor for clang at the moment, and we should use another library instead, like tbb or qt concurrent.

tcanabrava: Also note that for clang11 (unreleased) the omp reached maturity

patrickelectric: @tcanabrava clang supports openmp for a long time already, I don't know exactly when but it's available for us since the beginning. I did a quick test with clang6 and it works.

PR #911, opened 2020-03-16
Ping360 network configuration

patrickelectric:
Add Ping360 network configuration
Peek 2020-03-18 11-01


src/sensor/ping360asciiprotocol.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+
+QByteArray Ping360AsciiProtocol::staticIpAddressMessage(const QString& address)
+{
+    union Ipv4Union {
+        uint8_t values[4];
+        uint32_t ip;
+
+        // Return value with xxx format with leading zeros if necessary
+        QString getPing360IpValueFromIndex(int index) const
```
tcanabrava: I'v never seen a method within an Union. learning a thing every day.

src/sensor/ping360asciiprotocol.h
```diff
@@ -0,0 +1,86 @@
...
+ */
+class Ping360AsciiProtocol : public QObject {
+    Q_OBJECT
+public:
+    /**
+     * @brief The Discovery message used by the host computer to find Ping360.
+     *
+     * @return QByteArray
+     */
+    static QByteArray discoveryMessage();
```
tcanabrava: const static variable instead of static method?

patrickelectric: In pingviewer we usually do this kind of thing in structures, since they are data encapsulation. 
Ping360AsciiProtocol, as a class, uses functions, there is some methods that require arguments, and having messages that are not arguments but yes variables, will result in a confuse API, we should be consistent.

qml/DeviceManagerViewer.qml
```diff
@@ -212,6 +215,34 @@ PingPopup {
...
+                            PingImage {
+                                id: iconConfiguration
+                                source: StyleManager.settingsIcon()
+                                anchors.fill: parent
+                                selected: deviceConfigureMouseArea.containsMouse
+                                Layout.alignment: Qt.AlignRight
+
+                                visible: {
+                                    return connection.deviceType() == PingEnumNamespace.PingDeviceType.PING360 &&
+                                        connection.type() == AbstractLinkNamespace.Udp
```
Williangalvani: does it mean that the option will show when connected via the companion bridge? Can we also check if the port is 12345?

qml/Ping360EthernetConfiguration.qml
```diff
@@ -0,0 +1,118 @@
...
+            horizontalAlignment: Text.AlignHCenter | Text.AlignJustify
+            wrapMode: Text.Wrap
+            Layout.columnSpan: 2
+            Layout.fillWidth: true
+        }
+
+        PingTextField {
+            id: ipTextField
+            title: "IP address:"
+            text: "192.168.2.4"
```
Williangalvani: can/should we use the current ip address instead?

qml/Ping360EthernetConfiguration.qml
```diff
@@ -0,0 +1,118 @@
...
+            Layout.columnSpan: 2
+            Layout.alignment: Qt.AlignHCenter
+            visible: comboBox.currentIndex == 1
+            // TODO: Move to RegularExpressionValidator when using Qt 5.14+
+            // Check for valid IPV4 (0.0.0.0, 192.168.0.2)
+            validator: RegExpValidator {
+                regExp: /^(?!\.)((^|\.)([1-9]?\d|1\d\d|2(5[0-5]|[0-4]\d))){4}$/gm
```
Williangalvani: Nitpick: The annoying thing with this is that it doesn't let you erase all numbers before writing new ones. I wish it let you type but just didn't use the value instead... Do you think there is a better way?

qml/DeviceManagerViewer.qml
patrickelectric: The idea is to allow any UDP connection, why is there a need to check for the port ?

qml/Ping360EthernetConfiguration.qml
patrickelectric: Maybe @brianhBR can point which is the best static IP address for it.

qml/Ping360EthernetConfiguration.qml
patrickelectric: Can you share a gif ?  I was unable to replicate it.

patrickelectric: It'll not allow invalid values, if you start to type a invalid IP address, you'll stop you to typing.

patrickelectric:
> Another thing: We should warn the user if they are not on the same network as the Ping360 that the communication won't work, and ask them to set their IP address accordingly to the same network.

There is the warning message about it, it's also in the gif.

qml/DeviceManagerViewer.qml
Williangalvani: these messages will not get to the Ping360 if it is connected to the companion via serial and using bridges

Williangalvani: The gear icon will appear but won't work

qml/Ping360EthernetConfiguration.qml
Williangalvani: exactly, I just dislike that behavior, but I guess it is a Qt thing.

Williangalvani: if you have 192.168.2.4, you can't erase the 4 and then write a 5, you need to write a 5 and then erase the 4

qml/DeviceManagerViewer.qml
patrickelectric: No, that's not the case and it's not related to the UDP port but to the subnet network.
I'll add a warning message for it in the configuration.

patrickelectric: > these messages will not get to the Ping360 if it is connected to the companion via serial and using bridges

I got it now thanks!

qml/Ping360EthernetConfiguration.qml
Williangalvani: keep it as is, if anyone complains we can think about improving it

qml/DeviceManagerViewer.qml
patrickelectric: Done via check of isCompanion

qml/Ping360EthernetConfiguration.qml
brianhBR: I don't have any specific IP in mind as long as it is on the same network. I like the idea of warning the user if they pick something other than 192.168.2.xxx.

patrickelectric:
Added subnet check also.

src/link/linkconfiguration.cpp
```diff
@@ -11,6 +14,7 @@ const QMap<LinkConfiguration::Error, QString> LinkConfiguration::_errorMap {
     {InvalidArgsNumber, "Link configuration have a invalid number of arguments"},
     {ArgsAreEmpty, "Link configuration arguments are empty."},
     {InvalidUrl, "Url not formatted properly."},
+    {InvalidSubnet, "IP is not in a reachable network."},
```
Williangalvani:
```suggestion
    {InvalidSubnet, "IP is not in a reachable network. It may not respond to IP change messages. Consider adding your device to this network."},
```

patrickelectric: This is a generic message, should not be device specific.

Williangalvani: by "device" I meant computer

patrickelectric: "It may not respond to IP change messages." is a Ping360 specific thing, but I'll keep the final part.

patrickelectric: Moved to: "IP is not in a reachable subnet. Configure your computer network settings."

src/link/linkconfiguration.h
```diff
@@ -238,6 +240,14 @@ class LinkConfiguration : public QObject {
      */
     void setType(const LinkType& type) { _linkConf.type = type; };
 
+    /**
+     * @brief Check if IP is subnet
+     *
+     * @return true
+     * @return false if IP is not valid or is not an ethernet connection
```
Williangalvani:
```suggestion
     * @return false if IP is not in the same subnet as the computer or is not an ethernet connection
```

src/sensor/ping360helperservice.h
```diff
@@ -46,10 +62,24 @@ class Ping360EthernetFinder : public QObject {
      * @brief Stops broadcasting the discovery message
      *
      */
-    void stop();
+    void stopBroadcastService();
 
-signals:
+    /**
+     * @brief Configure device under specific IP and port as DHCP client
```
Williangalvani: port? is is in the IP:Port format?

patrickelectric: Oh, I thought that I have fixed this comments, thanks !

patrickelectric:
Travis will fail because transfer.sh is down, all tests are passing.

PR #916, opened 2020-04-13
firmware-update: Improve the instructions if the LED is blinking

patrickelectric:
There is no need to do a device recovery if the LED is blinking 
It's probably a hardware problem


patrickelectric:
@kklemens this steps is a way for the user to recover the firmware is something is wrong with it, we got a small number of people that reported that the LED was not blinking after opening it, we don't know why exactly.
The steps here is a help guide to flash the firmware if it does not have it already.

What I'm adding in this patch is to inform the user to talk with us if the LED is blinking after opening, I got a user that tried to do this procedure with the LED blinking, and this can make things worst.

kklemens:
Ok @danielgud FYI

PR #917, opened 2020-04-22
Add always-on-top mode and opacity configuration

patrickelectric:


qml/DisplaySettings.qml
```diff
@@ -50,6 +50,39 @@ ColumnLayout {
...
+                    text: "Opacity:"
+                }
+
+                ComboBox {
+                    id: opacityCB
+                    Layout.columnSpan: 4
+                    Layout.fillWidth: true
+                    currentIndex: SettingsManager.applicationOpacityIndex
+                    model: [100, 90, 80, 70]
+                    displayText: currentText + "%"
```
tcanabrava, line 79: nice trick

qml/DisplaySettings.qml
```diff
@@ -50,6 +50,39 @@ ColumnLayout {
...
+                Label {
+                    text: "Opacity:"
+                }
+
+                ComboBox {
+                    id: opacityCB
+                    Layout.columnSpan: 4
+                    Layout.fillWidth: true
+                    currentIndex: SettingsManager.applicationOpacityIndex
+                    model: [100, 90, 80, 70]
```
Williangalvani, line 78: I'd prefer this to be a slider (if that is not a bunch of unnecessary work)

patrickelectric: It was discussed in private

PR #932, opened 2020-10-14
Add option to run log with or without realtime update

patrickelectric:
Some people may turn on the sonar, connect with ping-viewer, but without request or starting the scanning with it, or stopping and starting ping requests, for the user to not wait this unnecessary amount of time when playing the log, it was added an option where the user will have the log processed with a maximum timeout of each message of 100ms. So, if any interval between each message is bigger than 100ms, the interface will not wait for the entire time and process the message inside the 100ms interval.

### DIFF
--- a/qml/DisplaySettings.qml
+++ b/qml/DisplaySettings.qml
@@ -97,6 +97,17 @@ ColumnLayout {
                     onCheckedChanged: SettingsManager.replayMenu = checked
                 }
 
+                CheckBox {
+                    id: realTimeReplayChB
+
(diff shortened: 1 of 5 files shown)


Williangalvani:
With 2m scan range, unchecking "realtime" makes the replay slower for me. Shouldn't it be normal unless a message takes over 100ms?

patrickelectric:
@Williangalvani Done, can you test with a log file to validate it ?

Williangalvani (approved):

PR #943, opened 2020-12-10
fixed naming bug in `Log.process`

Eliot-Insight:

PR #944, opened 2020-12-10
Process as generator; recover on error

Eliot-Insight:
- Implemented `__iter__` method of `Log` class as a generator to efficiently read `.bin` files without needing to store the entire file in memory (reads and yields one message at a time).
- Implemented corresponding `parser` method which is a generator that decodes the messages.
- Updated `if __name__ == '__main__'` script component to use the more efficient syntax.

### DIFF
--- a/examples/decode_sensor_binary_log.py
+++ b/examples/decode_sensor_binary_log.py
@@ -1,11 +1,18 @@
 #!/usr/bin/env python3
 
-import sys
-import struct
-from argparse import ArgumentParser
+import struct, sys, re
+
+# 3.7 for dataclasses, 3.8 for walrus (:=) in recovery
+assert (sys.version_info.major >= 3 and sys.version_info.minor >= 8), \
+    "Python version should be at least 3.8."
+
 from brping import PingParser
 from dataclasses import dataclass
 from typing import IO, Any
+
+
+def indent(obj, by=' '*4):
+    return by + str(obj).replace('\n', f'\n{by}')
 
 
 @dataclass
@@ -17,8 +24,7 @@ class PingViewerBuildInfo:
     os_version: str = ''
 
     def __str__(self):
-        return f"""
-PingViewerBuildInfo:
+        return f"""PingViewerBuildInfo:
     hash: {self.hash_commit}
     date: {self.date}
     tag: {self.tag}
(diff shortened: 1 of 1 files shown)


Eliot-Insight:
Includes the bug fix from #943, so if this is merged 943 can be removed/ignored.

Eliot-Insight:
Interesting that windows build failed - I haven't changed any of the actual ping-viewer code, just the example. I guess the master branch is currently failing the windows build too?

Williangalvani (commented):
Hi @Eliot-Insight ,

Python-wise this looks good. I'll have @patrickelectric look at the ping-specific part.

I found the "recover" method to be a little confusing to follow. Can you elaborate on the issue you are fixing? I tried manually corrupting a log by deleting some sections of it, but that made the script fail.

Eliot-Insight:
Hi @Williangalvani,

In my testing I found that when errors were occurring it was because the array-length that it was reading in was a huge number that tried to read all the remaining bytes of the file into a single message, which used up the rest of the file. I provided a couple of example files where this occurs [here](https://drive.google.com/drive/folders/1bSsMHbogu3H4xZU9hA9bcQFf1gmDFQgt?usp=sharing) (same as in my initial post). I'm unfortunately unsure how or why this was occurring in the first place, although those specific files were generated at a time where our communication was regularly dropping out because of some issues with our setup.

The idea with this recovery method is basically what I proposed in pseudocode in [my bluerobotics forum post](https://discuss.bluerobotics.com/t/ping360-parsing-recovery-issue-with-save-format-or-parser-script/8886/3?u=eliotinsight), where if it detects that a clearly incorrect array length has been read (`data_array_length > max_data_array_length`), it assumes that it's lost and tries to recover by finding the next available timestamp (recognised using regex), after which it parses that timestamp manually, and then reads the corresponding message normally, before returning to the standard read-timestamp, read-message process.

While the recovery process is intended (and seemingly works) for specifically the issue I was having, I'd imagine it should work for any time when bytes are randomly missing (although I haven't tested this), potentially at the cost of a few messages. If bytes get deleted in a file it'll at some point end up trying to read the next array length from a random position, which involves turning four random bytes into a 32-bit UINT, which will likely be long enough to trigger recovery mode, and if not would do so relatively soon afterwards as each incorrect length read will almost certainly take it to another incorrect location. 

If your manual corruption of the files was at the bit level then there's a 7/8 chance it's damaged the byte-alignment, in which case I would expect it to fail to recover after that point, as regex operates on bytes and won't find a valid timestamp location. If you deleted some bytes and it didn't work then possibly they were from the file header, but if not then I'm afraid I'm not sure why it's not working. If you can confirm that it's not working after some random bytes deletion from say the middle of a file I'll try to have a look into it in my spare time over the next week or so. Since it's working for my issue I can't justify spending more work time on it unfortunately.

Williangalvani (approved):
lgtm.
I wonder if the underlying issue is in Ping-Viewer or Ping360 itself, but this works for now.

We can look at random corruptions later, as they don't seem to be an issue.
could you squash the commits and prepend the commit message with "examples:"?

Eliot-Insight:
I've never squashed commits before, but it seems to have worked. Hopefully that's what you were after :-)

Williangalvani:
> I've never squashed commits before, but it seems to have worked. Hopefully that's what you were after :-)

Yes, that looks correct!

@patrickelectric do you mind taking a look at this?

patrickelectric (approved):

patrickelectric:
Hi @Eliot-Insight, sorry for the delay! Be free to ping us if you desire to get it merged asap.
Thank you very much for your contribution!

PR #951, opened 2021-02-22
Docs: add manual firmware update

jaxxzer:


docs/firmware-update.md
```diff
@@ -14,6 +16,24 @@ Steps to flash the device:
...
 
+## Manual Firmware Update
+
+1. [Connect](device-manager.md) to the device
+2. Open the Firmware Update menu
+3. Choose 'Manual Update'
+4. Download your desired firmware from the [firmware repository](https://github.com/bluerobotics/ping-firmware/tree/master/ping1d)
+  - Find your desired firmware in the `/ping1d` folder
+  - Click 'Raw' to view the raw document
+  - Right-click the webpage and click 'Save As...' to save the firmware hex file (`.hex`)
```
rafaellehmkuhl, line 27: This would be better if they were also an ordered list.

docs/firmware-update.md
```diff
@@ -2,6 +2,8 @@
 
 The Firmware Update menu allows you to update the programming on a Ping device.
 
+## Automatic Firmware Update
+
 Steps to flash the device:
```
rafaellehmkuhl, line 7: It seems like this should be outside (before) both subtopics ("Automatic" and "Manual Firmware Update").

docs/firmware-update.md
```diff
@@ -14,6 +16,24 @@ Steps to flash the device:
...
+  - Find your desired firmware in the `/ping1d` folder
+  - Click 'Raw' to view the raw document
+  - Right-click the webpage and click 'Save As...' to save the firmware hex file (`.hex`)
+5. In Ping-Viewer, click `Browse` to select the firmware hex file that you saved
+6. Click _FIRMWARE UPDATE_ to start the process.
+7. The update progress is indicated by the progress bar located at the bottom of the menu. When the update completes, the progress bar will resume it's idle animation, and the Ping device will resume data transmission.
+
+> Wait for the update process to complete before unplugging the device!
+
+![](/ping-viewer/images/firmware-update/manual.png)
```
rafaellehmkuhl, line 34: Missing image caption.

rafaellehmkuhl (changes requested):
Also, there are two images uploaded but only one of them ("manual.png") is used.

rafaellehmkuhl:
The travis error is unrelated to your PR. It's a problem with a dependency of the MKDocs.

docs/firmware-update.md
jaxxzer: done

docs/firmware-update.md
jaxxzer: doin

docs/firmware-update.md
jaxxzer: donezo

rafaellehmkuhl:
From my point of view, we can approve that PR without the third commit and move to another one to focus on the MKDocs build problem. I could also work on making MKDocs building on the other PR and we approve this one later. Let me know what you think @jaxxzer.

rafaellehmkuhl (approved):

PR #973, opened 2021-06-10
Fix flasher

patrickelectric:
Needs to test under windows and mac


Williangalvani:
This fixes the issue on my setup

PR #974, opened 2021-06-16
Docs Update

ES-Alexander:
- [x] add a hyperlink to the source code (github page) in the build instructions
- [x] clarify `.hex` download in firmware-update
- [x] some extra troubleshooting info
- [x] add Ping360 details/section on QuickStart
- [x] extra explanation for ping360 device/scan parameters 
(e.g. tuning and timing implications - similar to what's already done for the 'Range' parameter)


ES-Alexander:
Can we add a section to firmware-update about updating via companion? 

I've put a placeholder comment in the file (gets ignored in the generated html) - I'm suggesting we add a subsection to 'Manual Firmware Update'. Would likely be quite useful for people wanting to update firmware without needing to open their ROV (since currently we can't use ping-viewer to update with a UDP connection).

If this is ok to add I'm happy to if someone can tell me which port it would be available on. Would be good to also add this to the companion web interface docs as an available page.

ES-Alexander:
Not sure if better to merge now (pending response on 'companion' section in firmware-update), or add Ping360 details and extra info first.

docs/build-instructions.md
```diff
@@ -95,4 +92,5 @@ When installing, you must minimally select all Visual C++ components.
 5. Build the documentation.
     - `pipenv run mkdocs build`
 6. Access the generated documentation using your favorite browser.
-    - `cd ping-viewer && firefox index.html`
\ No newline at end of file
+    - `cd ping-viewer && firefox index.html`
+    - `cd ping-viewer && open -a "Firefox" index.html` on MacOS
```
patrickelectric, line 96: I believe that just `open` should do the job, not necessary firefox.

patrickelectric:
> I've put a placeholder comment in the file (gets ignored in the generated html) - I'm suggesting we add a subsection to 'Manual Firmware Update'. Would likely be quite useful for people wanting to update firmware without needing to open their ROV (since currently we can't use ping-viewer to update with a UDP connection).

I believe that is better to have such documentation in the companion documentation page, since this is out of the scope of ping-viewer and also could it could open ping-viewer documentation to include things that are not only ping-viewer related but ping-family ecosystem.

But, we could have a reference here in the documentation for the companion webpage. 



> Not sure if better to merge now (pending response on 'companion' section in firmware-update), or add Ping360 details and extra info first.

This already looks great! Thanks!
You can remove the commented topic about the companion update and we could merge ASAP.

You can work in the other topics in following PRs or finish here if you wish.

patrickelectric (approved):

docs/build-instructions.md
ES-Alexander: I was being consistent with the existing suggestion, but I suppose open uses the default browser which takes account of the user's preference, so fair point.

ES-Alexander:
> > I've put a placeholder comment in the file (gets ignored in the generated html) - I'm suggesting we add a subsection to 'Manual Firmware Update'. Would likely be quite useful for people wanting to update firmware without needing to open their ROV (since currently we can't use ping-viewer to update with a UDP connection).
> 
> I believe that is better to have such documentation in the companion documentation page, since this is out of the scope of ping-viewer 

Fair enough
> and also could it could open ping-viewer documentation to include things that are not only ping-viewer related but ping-family ecosystem.

Agreed. Any thoughts on where ping-family documentation should go? There are the individual product pages on the BR website but the family as a whole doesn't really have a place for documentation - the closest things would be this page and the ping-protocol one. 
> But, we could have a reference here in the documentation for the companion webpage.

Fair. That can be added once the companion webpage docs have a page for the ping page.

> > Not sure if better to merge now (pending response on 'companion' section in firmware-update), or add Ping360 details and extra info first.
> 
> This already looks great! Thanks!
> You can remove the commented topic about the companion update and we could merge ASAP.
> 
> You can also work in the other topics in following PRs or finish here if you wish.

I'll see what I can get done today. Will confirm within about 10 hours but should be good to merge after that. Anything left afterwards can be handled next week/later.

patrickelectric:
> Agreed. Any thoughts on where ping-family documentation should go? There are the individual product pages on the BR website but the family as a whole doesn't really have a place for documentation - the closest things would be this page and the ping-protocol one.

Maybe we could write or start a webpage for ping-family sensors, is there any topic that you find important for such thing exist ? The only thing common between all sensors from the ping-family is the protocol, but only that from my point of view.

ES-Alexander:
All done, added a bunch of Ping360 stuff in the various different pages :-)

I think it should all be ok, although someone should check the 'head down' description in display-settings (I can't remember if that's correct, and can't test it)

ES-Alexander:
> > Agreed. Any thoughts on where ping-family documentation should go? There are the individual product pages on the BR website but the family as a whole doesn't really have a place for documentation - the closest things would be this page and the ping-protocol one.
> 
> Maybe we could write or start a webpage for ping-family sensors, is there any topic that you find important for such thing exist ? The only thing common between all sensors from the ping-family is the protocol, but only that from my point of view.

Yeah, I'd agree that at this stage the only commonalities are:
- they're both sonars
- they share the same protocol (ish)
- they can both be used through companion

It probably wouldn't hurt to make a webpage for the family if it's likely to expand, but at this stage it's probably not worth it because it'd just be duplicating or linking to info that's already elsewhere. Probably better left for another day.

patrickelectric:
@ES-Alexander, the ping family sensors limit itself to the protocol, we have plans to have products that are not sonnars that speak ping-protocol.

patrickelectric:
Great contributions thanks!

PR #979, opened 2021-06-28
Updates for ping-viewer 2.2 release

patrickelectric:
Mostly interfaces changes and new information related to the newer version 
Some URL references was necessary to fix for the CI


patrickelectric:
@Eliot-Insight can you take a look ? For some reason I can't add you on the review request.

docs/display-settings.md
```diff
@@ -3,16 +3,25 @@
...
+The application will be over all windows.
+
+- **Opacity:**
+Select the application opacity.
 
 - **Enable Replay Menu:**
 Enable the [Replay Menu](replay-data.md)
 
+- **Enable Real-Time Replay:**
+Each profile message will be displayed as the real sensor and not periodically.
```
ES-Alexander, line 21: > Each profile message will be displayed as the real sensor and not periodically.

-> Shows playback as it was recorded (ie what you would see from a screen recording when the data was captured). Sets profile messages to display based on recorded timestamps, instead of using a constant period/update rate.

ES-Alexander (approved):
My comments are all just for additional clarity, so I'd recommend including them but that's a bit up to your judgement on whether the changes will actually be helpful to users. Otherwise all seems fine to me :-)

patrickelectric:
Thanks @Eliot-Insight !

patrickelectric:
Thanks @ES-Alexander !

PR #980, opened 2021-07-01
src: CMakeLists: Disable console on windows

patrickelectric:
Set program as a windows application


ES-Alexander:
Fixes #981

PR #986, opened 2021-07-12
docs: build-instructions: Add cmake build instruction for Ping-Viewer v2.2.0+

### LINKED ISSUE #985: docs: update build instructions for cmake
Current build instructions use qmake, which is no longer supported. 

@patrickelectric could you update them to display the relevant info for building with cmake instead? :-)

This was brought up in [this forum post](https://discuss.bluerobotics.com/t/building-pingviewer-from-source/10132)

patrickelectric:
Ping-Viewer from now on will use cmake build system.

Fix #985

### DIFF
--- a/docs/build-instructions.md
+++ b/docs/build-instructions.md
@@ -29,7 +29,7 @@ To get the source files:
 - **macOS:** v10.14 or higher
 - **Linux:** 64 bit, gcc-8 or clang-7 compiler versions or higher
 - **Windows:** 10 or higher, Visual Studio 2017 compiler (64 bit)
-- **Qt version:** 5.12 **(only)**
+- **Qt version:** 5.15 **(only)**
 
 > **Tip:** For more information see: [Qt 5 supported platform list](http://doc.qt.io/qt-5/supported-platforms.html).
 
@@ -42,7 +42,7 @@ When installing, you must minimally select all Visual C++ components.
 #### Install Qt
 
 1. Download and run the [Qt Online Installer](http://www.qt.io/download-open-source).
-2. In the installer *Select Components* dialog choose: 5.12.
+2. In the installer *Select Components* dialog choose: 5.15.
 
    Then install just the following components:
    - **Qt binary version:**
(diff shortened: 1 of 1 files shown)


docs/build-instructions.md
```diff
@@ -65,6 +65,22 @@ When installing, you must minimally select all Visual C++ components.
 
 #### Building with terminal
 
+Since Ping-Viewer v2.2.0, the project now uses CMake to simplify the build process.
+For more information about CMake with Qt based projects, check the [Qt CMake manual](https://doc.qt.io/qt-5/cmake-manual.html);
+
+1. Use CMake to configure the project
+    - `cmake -B build -DCMAKE_BUILD_TYPE=Debug`
```
ES-Alexander, line 72: Add a note to ensure Qt5 is on `PATH`?

ES-Alexander: Also make sure Qt5.15 is installed (current instructions are for 5.12). 

For Mac can use `brew install qt@5`, then `export PATH="/usr/local/opt/qt@5/bin:$PATH"` before doing `cmake -B build...`. Alternatively can permanently add to path by doing `echo 'export PATH="/usr/local/opt/qt@5/bin:$PATH"' >> ~/.zshrc`, then restarting terminal before building.

docs/build-instructions.md
```diff
@@ -65,6 +65,22 @@ When installing, you must minimally select all Visual C++ components.
...
+2. Use CMake again to build it
+    - `cmake --build build --parallel --config Debug`
+3. Go inside build folder
+    - `cd build`
+4. And after that you'll be able to run the project
+    - `./pingviewer`
```
ES-Alexander, line 78: Mac required `./pingviewer.app/Contents/MacOS/pingviewer` instead of `./pingviewer`.

By the way, I re-built after our discussion and the GUI elements work correctly now (e.g. it only auto-changes the distance while the box is checked, and the pings/s value isn't stuck at 0). Probably just an issue with how I was linking Qt - works with the PATH export I suggested above.

patrickelectric: I have added a note about macOS path.

docs/build-instructions.md
patrickelectric: Can you check the new instructions ? There a new tip about different installation environment. 
I also fixed a problem with the CMakeFile.txt for the project, so you should fetch the latest master branch now.
The problem appears to be with the Qt documentation where QtQuickCompiler configuration was necessary to use in a different way, the problem related to this module should not happen and the current PR should solve any problem related to the Qt5 search process.

patrickelectric:
@Eliot-Insight let me know if I can merge it

ES-Alexander (approved):
Building works fine now.
I'm still having issues with Ping1D, although that seems to be the case for even past PingViewer downloads - I think I've messed up my firmware or something now. Ping360 is working fine, so assuming it's likely ok, and the Ping1D issues are after building anyway, so would seem to be unrelated :-)

PR #992, opened 2021-08-03
Waterfall transparency fix

### LINKED ISSUE #933: Transparency gradient not working
Values such as #AARRGGBB are not working.

ES-Alexander:
Fix #933

Not yet tested, but I believe that's all that should be required (it just wasn't enabled/matched in the regex).


patrickelectric:
Hi @Eliot-Insight, I have added the necessary part that was missing for your PR, the github actions is running again, can you use the artifacts for there to test and see if the behavior is what we are looking for ?
From what I tested here it appears to be working now.

ES-Alexander (commented):
Works for me :D

patrickelectric (commented):
Thanks!

PR #1012, opened 2021-09-13
sensor/ping360helperservice: bugfix udp broadcast ip address assignemnt

jaxxzer:
The ping360 static ip address assignment was broken.

We were attempting to send the SetIP command directly to the sender. 
We need to broadcast the SetIP command on the same subnet where we received the Discovery response.

To test, assign a static ip to your computer's ethernet interface, then plug ping360 ethernet directly to your computer. Try to set your ping360 to a static ip address!


patrickelectric (approved):
Just need to fix the code style to pass by CI.

PR #1023, opened 2021-12-09
WIP: move to Qt6 

patrickelectric:


ES-Alexander:
Would be good to check for Ubuntu 18.04 compatibility - if this breaks it then it should be documented in whatever release notes we provide :-)

By the way, is this actually waiting on a review? Might be good to request that from someone else, and/or mark it as draft if it's not yet ready for merging.

patrickelectric:
@Eliot-Insight Sorry, I just moved it to a draft, it's a long term work to move everything to Qt6.

PR #1038, opened 2022-02-14
Fix parser problem using latest version of ping-cpp 

### LINKED ISSUE #1033: Re-Encoded Replay Display Issues
## Summary
Version: 2.3.1

Operating System: Mac (Monterey)

What is happening: As per PR #1029, where I implemented a binary log file encoder,

> I've tested the encoding and it seems to work properly (the output can be viewed by PingViewer), although sometimes the output doesn't display (may depend on how many/which messages are included).

How to reproduce it: 
1. Record a binary log with Ping Viewer
2. Decode it with Python
3. Remove/filter out some messages
4. Re-encode with #1029 
5. Try to re-play the re-encoded file with Ping Viewer (works fine with some re-encoded files, but SOMETIMES fails to display some others).

## Additional information

> **NOTE: @patrickelectric was not able to reproduce this with the same files**
> **NOTE 2: did some more testing, and found that sometimes the files work, but other times not**

https://user-images.githubusercontent.com/25898329/149761510-74741826-3a20-4f4e-a04f-fb42efaebd19.mov

GUI log from while I was taking the video (and unfortunately also from quite a while beforehand too I think):
[20220117-220504068.txt](https://github.com/bluerobotics/ping-viewer/files/7881477/20220117-220504068.txt)

I think the display inconsistency is most likely a Ping Viewer bug because I was writing only `2300` messages (and only `1300` messages for a Ping1D file), and depending on the start and stop points and step it would sometimes work and sometimes not. ~As I understand it, if the encoding works for one set of `2300` messages then it should also be fine with other ones - especially the same set of messages just with a few more/less included.~ Issue doesn't even seem to be entirely file based anymore, maybe just sometimes Ping Viewer fails to replay a file after another one, although it seems to fail more on some particular files (this seems like it's probably going to be pretty hard to track down...)

## Ping360
[ping360_F-fail_P-pass.zip](https://github.com/bluerobotics/ping-viewer/files/7767140/ping360_F-fail_P-pass.zip)

| test settings (Ping Viewer 2.3.1, and 2.3.0) | replay display works? |
| --- | --- |
| original file | ~no~ yes (just slow to start) |
| processed file (filter to just `2300` (`device_data`) messages) | yes |
| start at 20th `2300` message | yes |
| start at 30th `2300` message | yes |
| start at 20th `2300` message and skip every 2nd one | yes |
| start at 30th `2300` message and skip every 2nd one | ~no~ sometimes? |

## Ping1D
[ping1d_F-fail_P-pass.zip](https://github.com/bluerobotics/ping-viewer/files/7767246/ping1d_F-fail_P-pass.zip)

| test settings | replay display works? |
| --- | --- |
| original file | yes |
| processed file (filter to just `1300` (`profile`) messages) | ~no~ sometimes? |
| start at 2nd `1300` message | ~yes~ sometimes? |
| start at 10th `1300` message | ~no~ sometimes? |
| start at 20th `1300` message | ~no~ sometimes? |
| start at 30th `1300` message | yes |

patrickelectric:
Fix #1033

### DIFF
--- a/src/sensor/pingparserext.cpp
+++ b/src/sensor/pingparserext.cpp
@@ -5,12 +5,12 @@ void PingParserExt::clearBuffer() { _parser.reset(); }
 void PingParserExt::parseBuffer(const QByteArray& data)
 {
     for (int i = 0; i < data.length(); i++) {
-        PingParser::ParseState state = _parser.parseByte(data.at(i));
-        if (state == PingParser::ParseState::NEW_MESSAGE) {
+        PingParser::State state = _parser.parseByte(data.at(i));
+        if (state == PingParser::State::NEW_MESSAGE) {
             parsed++;
             _rxMessage = _parser.rxMessage;
             emit newMessage(_rxMessage);
-        } else if (state == PingParser::ParseState::ERROR) {
+        } else if (state == PingParser::State::ERROR) {
             errors++;
             emit parseError();
         }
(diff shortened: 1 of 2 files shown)


ES-Alexander (approved):
Seems to work - couldn't replicate the issue anymore :-)

PR #1045, opened 2022-05-05
protocoldetector: add default links for blueos

Williangalvani:
image

it did take ~50s to scan, though... plus it freezes after it is done, lol


Williangalvani:
after talking to @patrickelectric , I agree we should add a mdns resolver into ping-viewer itself to find devices with ping capabilities

PR #1057, opened 2023-04-07
qml: FirmwareUpdate: Sort firmware by version

### LINKED ISSUE #976: Firmware version for ping1D is not sort correctly
image

patrickelectric:
image
Fix #976


joaoantoniocardoso (approved):
I'm not used to QML, but the regex and logic make sense.

PR #1068, opened 2023-11-29
sensor: ping360: Improve test to check for firmware version over 3.3.0

patrickelectric:
The newer firmwares come with async mode

### DIFF
--- a/src/sensor/ping360.cpp
+++ b/src/sensor/ping360.cpp
@@ -15,6 +15,7 @@
 #include <QStringList>
 #include <QThread>
 #include <QUrl>
+#include <QVersionNumber>
 
 #include "hexvalidator.h"
 #include "link/seriallink.h"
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso (approved):
Cool!

src/sensor/ping360.cpp
```diff
@@ -99,19 +100,17 @@ Ping360::Ping360()
...
 
         // Wait for firmware information to be available before looking for new versions
         static bool once = false;
         if (!once && _commonVariables.deviceInformation.initialized) {
             once = true;
 
-            if (_commonVariables.deviceInformation.firmware_version_major == 3
-                && _commonVariables.deviceInformation.firmware_version_minor == 3
-                && _commonVariables.deviceInformation.firmware_version_patch == 1) {
+            if (version > QVersionNumber(3, 3, 0)) {
```
joaoantoniocardoso, line 113: Wait, we changed here from 3.3.1 to 3.3.0!

joaoantoniocardoso: hmm, it was `==`, not `>=`. Is this okay?

patrickelectric: That's the intended behavior, the async feature is only available on newer firmware, that are newer than 3.3.0, it started on 3.3.1

joaoantoniocardoso: okay

joaoantoniocardoso (approved):

